Issue #78, opened 2023-03-27
Nested dictionaries or lists as Fractal Task argument inputs


jluethi:
Following up on https://github.com/fractal-analytics-platform/fractal-web/issues/65

> It should be established if the kind of elements within the list will always be objects or if the client could expect to find either string/number/boolean as list elements.

At the moment, the only use-case we have for lists are lists of dictionaries, see Create OME-Zarr task example below. It doesn't really make sense to have a list of key-value pairs that are strings (=> that would be a dictionary). But I can't exclude that we would eventually come up with tasks where the lists contain strings directly, or even other lists? But the priority would be list of dictionaries.

Currently, one can import a workflow that contains a list of dictionaries, but the interface then doesn't show those objects anymore (see https://drive.google.com/file/d/1fi4DIY1_3x8EhIYeLe9XSaysd-VstZkA/view?usp=sharing where we have a typical Create OME-Zarr task => would be a list of dictionaries, each dictionaries containing string or number inputs). This would be a typical input:

```
{
    "allowed_channels": [
        {
            "colormap": "00FFFF",
            "wavelength_id": "A01_C01",
            "end": 800,
            "label": "DAPI",
            "start": 110
        },
        {
            "colormap": "FF00FF",
            "end": 290,
            "wavelength_id": "A01_C02",
            "label": "nanog",
            "start": 110
        },
        {
            "colormap": "FFFF00",
            "end": 1600,
            "wavelength_id": "A02_C03",
            "label": "Lamin B1",
            "start": 110
        }
        ]
    ,
    "coarsening_xy": 2,
    "num_levels": 5,
    "image_extension": "png"
}
```

What we certainly will need to support are nested dictionaries. Here are the 2 main examples of those. As you can see in the second one, we can have up to 3 levels of dictionaries at the moment

Examples of nested dictionaries:
Illumination Correction:
```
{
    "overwrite": "True", 
    "dict_corr": {
        "root_path_corr": "/data/homes/jluethi/v1_fractal/fractal-demos/examples/04_cardio_illumination_correction/../illum_corr_images/", 
        "A01_C01": "20220621_UZH_manual_illumcorr_40x_A01_C01.png", 
        "A01_C02": "20220621_UZH_manual_illumcorr_40x_A01_C02.png", 
        "A02_C03": "20220621_UZH_manual_illumcorr_40x_A02_C03.png"
        }
}
```

Example of double-nested dictionaries:
```
{
    "level": 0, 
[...]

Issue #77, opened 2023-03-27
Allow edits of `WorkflowTask.meta` field


tcompa:
There was a lot of progress in how to edit the task arguments (ref #65 #63), but the same feature should also be available for the `meta` field.

I propose we implement it as simply as possible, for the moment, that is by assuming that the `meta` field is a key-value object without nesting (that is, keys are string and values are string/numbers/booleans).

This will for sure cover current use cases (where we typically don't even set `WorkflowTask.meta`), and it will likely be enough also for the ones that will be introduced soon (ref https://github.com/fractal-analytics-platform/fractal-server/pull/582).

I'd say let's postpone the discussion on possibly more complex cases.

The following features are supported in fractal-server, and they should be also available in fractal-web:
1. Adding a new key-value pair (provided the key name is not already present);
2. Modifying the value in a given key-value pair;

(In the future we may add the possibility of removing a single key-value pair, or of replacing the whole `WorkflowTask.meta` field with a `null` value, but those features are not there yet)

PR #81, opened 2023-03-29
Arguments and Metadata of workflow tasks handling

rkpasia:
This work should close:
- #78 
- #77

A component structure has been created to handle every possible combination of `key-value` within `args` and `meta` properties of a workflow task.

Now, associated to a key, we could have either a `object` | `array` | `string` | `number` | `boolean`.

The logic is all there, there may be room for improvements UX-side.


src/lib/components/workflow/ArgumentForm.svelte
rkpasia, line 1: As one can see, all the cluttered logic that this component had before has been broken-down to the components structure within the `/workflow/common` components directory.

tcompa:
Nice, thanks.

A minor glitch: the "mylist-list-item" label should probably be "item":
Screenshot from 2023-03-29 12-04-56

Issue #25, opened 2023-03-02
Extract expiration time directly from JWT token


tcompa:

rkpasia:
This should be closed by 07696720

In order to solve this issue I've introduced a project dependency.
Specifically the `jose` npm package that allow to decode JWT token claims.

Now the clien't nodejs server receives the auth token through HTTP auth request. Once the token is received, the `exp` claim is retrieved and the `expiration` attribute of the `Set-Cookie` response header (used to indicate the browser to set a new cookie) is set accordingly to it. This allow the client to set a cookie with an expiration that is the same as the one set by the fractal-server.

PR #130, opened 2023-04-26
Sveltekit server api proxy

tcompa:
In progress

(@rkpasia, I'm opening the PR so that I get notified about new commits - no other reason)

Checks to mark progress:

- [x] Projects page
  - [x] List projects
  - [x] Delete project
  - [x] Create project
- [x] Single project page
  - [x] List datasets on load
  - [x] Create dataset
  - [x] Delete dataset
  - [x] List workflows on load
  - [x] Create workflow
  - [x] Import workflow
  - [x] Delete workflow
  - [x] Update project properties
- [x] Dataset page
  - [x] Load dataset
  - [x] Update dataset properties
  - [x] Create new dataset resource
  - [x] Delete dataset resource
- [x] Workflow page
  - [x] Load workflow data
  - [x] Update workflow properties
  - [x] Add new workflow task
  - [x] Delete workflow task
  - [x] Update workflow task args & meta
  - [x] Reorder workflow tasks
  - [x] Export workflow
  - [x] Apply workflow
- [x] Tasks page
  - [x] Load/List tasks
  - [x] Refresh tasks list
  - [x] Create single task
  - [x] Collect task
- [x] Authentication
  - [x] Login
  - [x] Redirect after successful login
  - [x] Logout

### DIFF
--- a/src/hooks.server.js
+++ b/src/hooks.server.js
@@ -1,20 +1,25 @@
 import { FRACTAL_SERVER_HOST } from '$env/static/private'
 
 export async function handle({ event, resolve }) {
+	console.log(`[${event.request.method}] - ${event.url.pathname}`)
 
 	if (event.url.pathname == '/' || event.url.pathname.startsWith('/auth')) {
+		console.log('Public page - No auth required')
 		return await resolve(event)
 	}
 
 	// Authentication guard
 	const fastApiUsersAuth = event.cookies.get('fastapiusersauth')
 	if (!fastApiUsersAuth) {
+		console.log('Authentication required - No auth cookie found - Redirecting to login')
 		return new Response(null, { status: 302, headers: { location: '/auth/login' } })
 	}
+
 	const whoami = await event.fetch(`${FRACTAL_SERVER_HOST}/auth/whoami`)
 	if (whoami.ok) {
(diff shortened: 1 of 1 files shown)


tcompa:
Some points to keep in mind, as part of the ongoing refactor. Comments concern the current version (https://github.com/fractal-analytics-platform/fractal-web/pull/130/commits/c0e249c2fb2f8361715c03e63c187308930fea00).

1. At the moment we rely on a `fetchContext` function, exported in `src/lib/server/common/fetchContext.js`. Could we simply export a given `fetch` function in a similar (tiny, single-purpose) file (either the native one or the SvelteKit one), and import it every time it's needed?
Question: when we use SvelteKit `fetch` function as an argument of (e.g.) a `load` function, is it dressed with additional attributes that are not available if we just import it?
2. We are now hard-coding the cookie expiration time, which is not great because it needs to match the one in fractal-server. Can we extract it automatically? Ref #25
3. Again about the cookie: what about non-localhost addresses (e.g. 127.0.0.1)? What's the strategy there? Should we move towards domain definition in apache (including for the fractal-server instance).
4. In v0.3, we have introduced stores in order to reduce the number of API calls. How does that change when moving to server-side rendering?

PR #136, opened 2023-05-16
Align with Fractal 1.3

tcompa:
I started updating all endpoint paths/query/body parameters and function signatures (when needed), but did not modify the actual function calls.

The following functions were modified (i.e. I added `projectId` to their arguments), and their use needs to be updated:
* getJob
* getWorkflow
* updateWorkflow
* deleteWorkflow
* exportWorkflow
* createWorkflowTask
* updateWorkflowTaskArguments
* updateWorkflowTaskMetadata
* deleteWorkflowTask
* downloadWorkflowJobLog


rkpasia (approved):

rkpasia:
I think this PR is ready to go.
We should add some automatic testing, but it will come.

tcompa:
> I think this PR is ready to go.

Thanks. I'm merging this one to unlock #138, and then I'll keep testing.

PR #184, opened 2023-06-19
Update package.json (address sweetalert issue)

tcompa:
Address:
* https://github.com/fractal-analytics-platform/fractal-web/security/dependabot/6. This only required an update of `package-lock.json`
* https://github.com/fractal-analytics-platform/fractal-web/security/dependabot/1 (this is not a real vulnerability, but let's keep it away from the repo anyway so that we don't get annoying alerts). This requires a downgrade of sweetalert, also in package.json`


rkpasia (approved):
I don't see any point against this.

PR #188, opened 2023-06-20
Integrate JSchema component within Workflow page

rkpasia:
This integration enables the usage of JSchema component as mean to handle arguments edits with schema restrictions.


rkpasia:
This solves:
- #190 
- #195
- Failed validation of the args data received from the server must only raise a warning, and not an error
- Always validate data upon "Save", and raise an actual error if this fails (with logs visible to the user, not only in the browser console)
- "Discard updates" should bring back the original data (i.e. the ones received from the server, ~~or from the sandbox prompt~~)

It's ready to be merged.

Issue #201, opened 2023-06-22
Display key, for key-value items in object arguments


tcompa:
I confirm that https://github.com/fractal-analytics-platform/fractal-web/issues/176 works as expected, and I can add/remove key-value pairs.
One improvement which is probably required is to also show the key name, which is currently not visible:
![Screenshot from 2023-06-22 08-51-30](https://github.com/fractal-analytics-platform/fractal-web/assets/3862206/f72116a6-f93a-43e7-8324-d946431beb2e)

Issue #198, opened 2023-06-21
For arguments defined via `definitions`, show the correct description


tcompa:
We have the following JSON Schema
```json
{
  "title": "CellposeSegmentation",
  "type": "object",
  "properties": {
    ...
    "channel": {
      "$ref": "#/definitions/Channel",
      "description": "Primary channel for segmentation; requires either ``wavelength_id`` (e.g. ``A01_C01``) or ``label`` (e.g. ``DAPI``)."
    },
    "channel2": {
      "$ref": "#/definitions/Channel",
      "description": "Second channel for segmentation (in the same format as ``channel``). If specified, cellpose runs in dual channel mode.  For dual channel segmentation of cells, the first channel should contain the membrane marker, the second channel should contain the nuclear marker."
    },
    ...
  },
  "required": [
    "input_paths",
    "output_path",
    "component",
    "metadata",
    "level",
    "channel"
  ],
  "additionalProperties": false,
  "definitions": {
    "Channel": {
      "title": "Channel",
      "description": "A channel which is specified by either ``wavelength_id`` or ``label``.",
      "type": "object",
      "properties": {
        "wavelength_id": {
          "title": "Wavelength Id",
          "type": "string"
        },
        "label": {
          "title": "Label",
          "type": "string"
        }
      }
    }
  }
}
```
Where we have two arguments `channel` and `channel2` which are of the same `Channel` type.

We currently display the `description` field of `Channel` (from the `definitions` block), rather than the actual `channel` and `channel2` descriptions. I think this is wrong, because there is then no way to identify what should go into `channel` and what should go into `channel2`. Can we show the argument description there @rkpasia?

I understand that we would not be showing some information (i.e. the string `"A channel which is specified by either ``wavelength_id`` or ``label``."`), but that seems less important.

tcompa:
Note that things may get more complex when we have `allowed_channels: list[OmeroChannel]` or even `allowed_channels: dict[str,list[OmeroChannel]]`, but at least for the case described above I have a strong opinion on the "correct" behavior.

tcompa:
For the record, as an example, this is with 28a9fb8aab9a93f58fc3a1ff7c058d253ef7d59d:

![Screenshot from 2023-06-23 12-47-05](https://github.com/fractal-analytics-platform/fractal-web/assets/3862206/0b6316b1-83c1-4e41-8a74-105316f754da)

where only channel has a description, which is the one in `definitions` entry.

[...]

PR #220, opened 2023-06-26
Show custom property key as title 

rkpasia:


rkpasia:
This could also be a fix for #198. I'm not able to reproduce a meaningful example for #198. Need help @tcompa

tcompa:
Thanks @rkpasia, this is indeed the expected behavior for #201, since the key name is displayed on top of the value (e.g. "MyKey1", in the screenshot).

Screenshot from 2023-06-26 10-26-24

tcompa:
As for #198, the behavior clearly improved, since we now see different descriptions for `channel` and `channel2` - see below.
What is currently wrong is that the titles are still not correct. They both display `Channel`, but they should match to the first-level argument names (as for the docstring), which are `channel` and `channel2`.

Screenshot from 2023-06-26 10-31-01

rkpasia:
In order to see "channel2" as title for the property, it should be included in the nested schema. I mean:

Given the following partial schema
```json
"channel2": {
      "$ref": "#/definitions/Channel",
      "description": "Second channel for segmentation (in the same format as ``channel``). If specified, cellpose runs in dual channel mode.  For dual channel segmentation of cells, the first channel should contain the membrane marker, the second channel should contain the nuclear marker."
    },
```

If no `title` property is present in the inner `channel2` schema, then the default title in the referenced schema `#/definitions/Channel` is used.

Instead, if a `title` property is present in the inner schema as in the following
```json
"channel2": {
      "$ref": "#/definitions/Channel",
      "title": "Channel2",
      "description": "Second channel for segmentation (in the same format as ``channel``). If specified, cellpose runs in dual channel mode.  For dual channel segmentation of cells, the first channel should contain the membrane marker, the second channel should contain the nuclear marker."
    },
```

then the client will display "Channel2" as property title/name

tcompa:
I see, thanks. Then this is a bit more work on the schema-generation side.

rkpasia:
I could indeed show the property key `channel2` instead of the `title` property. But that isn't really by Json Schema design.

Moreover, this would require more logic to discriminate the property title:
- if there is a title property, then we should display the title
- if there is no title inside the root schema, but there is a reference, should the client display the title or the key?
- There would be situations in which is more helpful to display the key or the title?
- [...]

tcompa:
> I could indeed show the property key channel2 instead of the title property.

I don't think fractal-web should do this. This is some more "custom" logic, which we will include in our schema-generation tools (in fractal-tasks-core). If a title is missing (as they are now), I'm OK with fractal-web not making any smart guess.

tcompa:
Here are three relevant screenshots with comments

Screenshot 1: all as expected

Screenshot from 2023-06-26 11-48-03

Screenshot 2: all as expected. The argument descriptions may be improved (the higher-level one points to the lower-level one, and the lower-level one points to the higher-level one), but this has nothing to do with fractal-web.

Screenshot from 2023-06-26 11-48-30

Screenshot 3: all as expected, apart from a glitch (the "undefined" below `CHANNEL_KEY`). Is this something we can still fix @rkpasia?
Screenshot from 2023-06-26 11-48-51

rkpasia:
> Screenshot 3: all as expected, apart from a glitch (the "undefined" below CHANNEL_KEY)


[af2d6e1](https://github.com/fractal-analytics-platform/fractal-web/pull/220/commits/af2d6e1e58060c74ea7df2a62e84a917eef99551) Should check it

jluethi:
Looks good to me!

tcompa (approved):

PR #226, opened 2023-06-27
Fix list items default collapsed accordion

rkpasia:


jluethi (approved):
Have you tested the behavior in Fractal web itself? Haven't started a local server with it, but the code change looks very reasonable to me!

rkpasia:
Yes, tested locally!

PR #228, opened 2023-06-27
Improve layout of workflow task panel

rkpasia:


rkpasia:
I've moved the change tracking buttons to the bottom of the panel. I think that this is a reasonable solution. Moving them to the tabs header would place them in an out-of-context position.

Feedbacks? @tcompa @jluethi 

Screenshot 2023-06-27 alle 11 57 37

tcompa:
> I've moved the change tracking buttons to the bottom of the panel. I think that this is a reasonable solution. Moving them to the tabs header would place them in an out-of-context position.
> Feedbacks? @tcompa @jluethi

Looks good to me

jluethi:
Hmm, that seems risky that users then just will forget to save.
I just noticed that we let users leave the page after they made un-saved edits and, if they return, their changes are discarded.

If we enforce a save or discard decision (as sketched in the figma board here: https://www.figma.com/file/RAivaBmlX8yzu7aEAwrxaX/Fractal-UI-Board?type=whiteboard&node-id=258-488&t=OKUQpf7TvG5XlVhh-0)
Screenshot 2023-06-27 at 12 05 58


Then I'd be fine with this change :)

jluethi:
Also tested this. I like the more compact header!

Also nice how the save dialog is always shown, even if the list needs to be scrolled. With that, I think we could even merge this before #230 

We really need to address #230 independent of this design choice though!

rkpasia:
I'll merge this one and open another PR just related to #230.

PR #233, opened 2023-06-28
Improve management of tasks within the client

### LINKED ISSUE #149: Display task list with richer structure
- [x] Change order of tasks in the `/tasks` page. Sorting should be according to `task.owner` (first the ones with `null`, then the others), then `task.name`, then `task.version` (decreasing)
- [ ] Change menu to select task to be added to a workflow, as in https://github.com/fractal-analytics-platform/fractal-web/issues/149#issuecomment-1569616877 and https://github.com/fractal-analytics-platform/fractal-web/issues/149#issuecomment-1609161737

rkpasia:
The user should be able to navigate through the tasks in the client with more context.

That is:
- Provide better sorting of tasks in tasks page.
- Provide better filtering of tasks in workflow-building page.

Merging this PR should close #149

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
 				"ajv": "^8.12.0",
 				"jose": "^4.14.4",
 				"json-schema-traverse": "^1.0.0",
+				"semver": "^7.5.3",
 				"sweetalert2": "^11.4.8"
 			},
 			"devDependencies": {
@@ -2595,7 +2596,6 @@
 			"version": "6.0.0",
 			"resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-6.0.0.tgz",
 			"integrity": "sha512-Jo6dJ04CmSjuznwJSS3pUeWmd/H0ffTlkXXgwZi+eq1UCmqQwCh+eLsYOYCwY991i2Fah4h1BEMCx4qThGbsiA==",
-			"dev": true,
 			"dependencies": {
 				"yallist": "^4.0.0"
 			},
@@ -3363,10 +3363,9 @@
 			}
 		},
 		"node_modules/semver": {
-			"version": "7.5.1",
-			"resolved": "https://registry.npmjs.org/semver/-/semver-7.5.1.tgz",
-			"integrity": "sha512-Wvss5ivl8TMRZXXESstBA4uR5iXgEN/VC5/sOcuXdVLzcdkz4HWetIoRfG5gb5X+ij/G9rw9YoGn3QoQ8OCSpw==",
-			"dev": true,
+			"version": "7.5.3",
+			"resolved": "https://registry.npmjs.org/semver/-/semver-7.5.3.tgz",
+			"integrity": "sha512-QBlUtyVk/5EeHbi7X0fw6liDZc7BBmEaSYn01fMU1OUYbf6GPsbTtd8WmnqbI20SeycoHSeiybkE/q1Q+qlThQ==",
 			"dependencies": {
 				"lru-cache": "^6.0.0"
 			},
@@ -4195,8 +4194,7 @@
 		"node_modules/yallist": {
 			"version": "4.0.0",
 			"resolved": "https://registry.npmjs.org/yallist/-/yallist-4.0.0.tgz",
-			"integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A==",
-			"dev": true
+			"integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A=="
 		},
 		"node_modules/yaml": {
 			"version": "2.2.2",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
 		"ajv": "^8.12.0",
 		"jose": "^4.14.4",
 		"json-schema-traverse": "^1.0.0",
+		"semver": "^7.5.3",
 		"sweetalert2": "^11.4.8"
 	}
 }
--- a/src/routes/tasks/+page.svelte
+++ b/src/routes/tasks/+page.svelte
@@ -2,6 +2,7 @@
 	import { onMount } from 'svelte';
 	import { page } from '$app/stores';
 	import { enhance } from '$app/forms';
+	import semver from 'semver';
 	import { collectTaskErrorStore } from '$lib/stores/errorStores';
 	import { taskModal as taskModalStore } from '$lib/stores/taskStores';
 	import TaskInfoModal from '$lib/components/tasks/TaskInfoModal.svelte';
@@ -19,7 +20,45 @@
 		if (error) setErrorReasons(error);
 	});
 
-	onMount(async () => {});
+	onMount(async () => {
+	});
+
+	// Sort tasks by owner, by name and by version
+	tasks.sort((t1, t2) => {
+		if (t1.owner === null) {
+			// If t1 owner is null, t1 should go before t2 if t2 owner is null
+			// Should check if t2 owner is null too
+			if (t2.owner === null) {
+				// Both owners are null, sort by name
+				if (t1.name < t2.name) return -1;
+				if (t1.name > t2.name) return 1;
+				// Names are equal, sort by version
+				const t1Version = semver.coerce(t1.version);
+				const t2Version = semver.coerce(t2.version);
+				const t1VersionLt = semver.lte(t1Version, t2Version);
+				if (t1VersionLt) return -1;
+				if (!t1VersionLt) return 1;
+			} else {
+				// t1 owner is null, t2 owner is not null, t1 should go before t2
+				return -1;
+			}
+		} else {
+			// t1 owner is not null, t2 owner is null, t2 should go before t1
+			if (t2.owner === null) return 1;
+			// Both owners are not null, sort by owner
+			if (t1.owner < t2.owner) return -1;
+			if (t1.owner > t2.owner) return 1;
+			// Owners are equal, sort by name
+			if (t1.name < t2.name) return -1;
+			if (t1.name > t2.name) return 1;
+			// Names are equal, sort by version
+			const t1Version = semver.coerce(t1.version);
+			const t2Version = semver.coerce(t2.version);
+			const t1VersionLt = semver.lte(t1Version, t2Version);
+			if (t1VersionLt) return -1;
+			if (!t1VersionLt) return 1;
+		}
+	});
 
 	function setErrorReasons(value) {
 		errorReasons = value;
@@ -174,7 +213,7 @@
 						<tr>
 							<td class="col-3">{task.name}</td>
 							<td class="col-1">{task.version}</td>
-							<td class="col-1">{task.owner}</td>
+							<td class='col-1'>{task.owner || "–"}</td>
 							<td class="col-2">
 								<button
 									data-fc-task={task.id}


tcompa:
Current status of the tasks page already looks good:

Screenshot from 2023-06-28 12-11-37


I can review more in detail later, especially if this has to evolve a bit more.

(minor decision: should versions be decreasing or increasing?)

tcompa:
@rkpasia note that I included multiple versions of the same package, for testing - see https://github.com/fractal-analytics-platform/fractal-web/commit/e9322ea5fdb674f7c482c5de89d1223e07a7e241

rkpasia:
Great! I also tested by editing the database entries 👌🏼

__tests__/semver.test.js
```diff
@@ -22,4 +23,31 @@ it('should be able to loosely validate versions', () => {
...
+
+	const versions = [
+		{ version: '0.10.0c0' },
+		{ version: '0.10.0b4' },
+		{ version: '0.10.0alpha3' },
+		{ version: '0.10.0a2' },
+		{ version: '0.10.0a0' },
+		{ version: '1.0.0rc4.dev7' },
+		{ version: '0.10.0beta5' },
+		{ version: '0.10.0alpha0' }
```
tcompa, line 38: We should also include `0.10.0` here @rkpasia

tcompa:
Testing with 232dd05e635882da0d52b76403fdd1b0db085118, and also the order of prerelease versions is now correct:

Screenshot from 2023-06-30 12-41-36

tcompa:
For the record, @rkpasia introduced the modal to add a task to a workflow as a placeholder. This is not (yet) meant to be working/testable.

jluethi:
Looking promising. Ping me once you want me to test this as well :)

jluethi:
😍 @rkpasia the new task page is awesome! That makes it much more structured, easy to keep an overview!

The task ordering seems to work as I'd expect it and it selects the 0.10.0 over 0.10.0a6 correctly!

Screenshot 2023-06-30 at 17 36 07

The split into the tabs for common tasks & User tasks also looks nice! I haven't tested it in the multi-user case yet, will follow up on that in a bit.

----

An important thing that I'd still add now would be how to handle custom tasks without (unique) versions. If one adds a task via the web interface, so far the user can't specify a version (we should eventually also fix that, but not so urgent). It then gets displayed as follows (2 custom tasks with the same name, but different source):

Screenshot 2023-06-30 at 17 26 01

That's quite suboptimal, as there's no way to know which one is which.
This would also happen if the user provides a version, but that version is not unique (we're currently not enforcing unique versions). What we are enforcing are unique sources. Thus, let's proceed as suggested here:
https://github.com/fractal-analytics-platform/fractal-web/issues/149#issuecomment-1609265339

If a given user has multiple tasks with the same name, we show a drop-down.
There, following your logic:

> If a group of users' tasks have same name+owner and they all have different versions, proceed as for common tasks (i.e. display them as grouped).
> If the versions are not unique, we could do dropdown by source. Alternatively, we just always to dropdown by source for user specific tasks with multiple tasks of the same name. Now source is not as elegant as version of course, but it will always be unique.

rkpasia:
In the changes introduced, [54a0a10](https://github.com/fractal-analytics-platform/fractal-web/pull/233/commits/54a0a1097c0ca612af34ae5335842c4d22f08d6e) enables 

> We could order the user task list in the following way:
Logged-in user comes top, afterwards alphabetical by task.owner [=> username / slurm_user (users need to have one of those 2 set to register custom tasks)].
And the task would be shown as follows (if logged in as jluethi):
jluethi: Joel_task
jluethi: TestTask
admin: TestTask
joel_2: Joel_2_Task
> But if one is logged in as admin, one sees:
admin: TestTask
jluethi: Joel_task
jluethi: TestTask
joel_2: Joel_2_Task

[14d204d](https://github.com/fractal-analytics-platform/fractal-web/pull/233/commits/14d204d0c6f0e63faa4292ddb71664da7be0e8f7) introduces a project dependency that I selected from npm which is [slim-select](https://www.npmjs.com/package/slim-select).
This library enables us to group and label many tasks together specifying how we want them to be grouped by as suggested
> And a word regarding ordering of common task: Ordering by package should be the goal. I think the current sorting achieves this. But if we change sorting, let's explicitly order the common task by their package :)

> Alternatively, is it possible to have non-selectable "headers" in the dropdown? For packages in the common case, for username in the other case.

Screenshot 2023-07-04 alle 08 27 13
Screenshot 2023-07-04 alle 08 27 02

src/lib/components/workflow/WorkflowTaskSelection.svelte
```diff
@@ -36,17 +56,49 @@
...
-	function setSelectedGroup(group) {
-		console.log(group);
-		selectedTypeOfTask = group;
+	function setSelectionControlData() {
+		let optionsMap = [];
+
+		if (selectedTypeOfTask === 'common') {
+			// Group filtered tasks by source
+			optionsMap = filteredTasks.reduce((dataOptions, task) => {
+				const source = task.source.split(':')[1]; // Package name
```
rkpasia, line 68: To get the package a task belongs to, I split its source property and take the second (1) element of the split string.
Is it sound enough for every case? @tcompa

rkpasia: Could it happen that a task has no source or edge cases to consider?

jluethi: I'll let Tommaso analyse this in detail. My understanding is that common tasks will always have a very predictable source structure, but user tasks could have any string as source, as long as it's unique.

jluethi:
This looks super cool! I'm collecting a few version of task packages & will install some custom tasks to review this.

jluethi:
@rkpasia 
The new layout for common task per package in the dropdown is great, sorting of versions works as expected. Let's go with that!

The sorting of tasks by users also looks great. Can we take always put the currently logged in user top or does that get too complicated? Current sorting is ok, this sorting with logged in user at the top (if they have tasks) would be even better

Search is not functional yet, right? But great idea, would love search when adding the tasks!

And for multiple user tasks with the same name, but no version, we still get the dropdown of all "Not specified" => if that would be replaced by the source in 2 cases, then it would work best.
The 2 cases are:
1. No version is present (currently gets replaced with "Not specified")
2. Version is not unique for that user & task

src/lib/components/workflow/WorkflowTaskSelection.svelte
tcompa: This is all sound.

By construction, our common tasks come from the installation&collection of a Python package, and the source looks like `pip_remote:fractal_tasks_core:0.9.0:::cellpose_segmentation` or `pip_remote:fractal_tasks_core:0.10.0:fractal-tasks::copy_ome-zarr_structure`, where the second field is indeed the package name.

Non-common tasks, on the contrary, only have a single constraint on the source, which is that the first field should match with `task.owner`.

These structures for the task sources are not strictly enforced, but they will always be there "by construction". Therefore they will hold until we make significant changes to the API (e.g. if we wanted to allow a change of owner for a given task, then it wouldn't be obvious what the source should be).

What is guaranteed is that all tasks have a source, and that this source has to be unique.

TL;DR
@rkpasia you are relying on very reasonable assumptions. They could evolve in the future, but that would only come with new use cases which we are currently not supporting.

tcompa:
I confirm that the behavior for common tasks looks reasonable to me:

Screenshot from 2023-07-04 10-35-54

Screenshot from 2023-07-04 10-36-02

tcompa:
> Search is not functional yet, right? But great idea, would love search when adding the tasks!

Agreed, that would be a nice idea. What would we be searching though? Task name?
Anyway, if it's not currently functional let's just remove it for the moment, and we can enable it later.

jluethi:
If it's a trivial part of the framework to allow search by task name, that would make things much simpler once there are many task.
i.e. I named my task "custom_measurement_joel", I don't see it in the list right away, so I start typing:
custom... => only see the task where the names contain the word custom.

Any fancier search can be postponed without issues. And if it's tricky to get to the core search pattern, we can also postpone that (and thus remove the field for the moment).

rkpasia:
I still have to read through the documentation of `slim-select` to understand how the search box works.

I think that in principle should be useful to search by task name. Don't know if it could be possible to also introduce a filtering by specifying, for instance, the package name or the package version. Maybe applying a custom search function could be doable.

jluethi:
> I think that in principle should be useful to search by task name.

Agreed, this will be very useful!

> Don't know if it could be possible to also introduce a filtering by specifying, for instance, the package name or the package version. Maybe applying a custom search function could be doable.

Given the need for documentation and testing for Fractal web, I'd say a short effort on search by task name is still in scope, custom search functions to include package names, versions etc. is out of scope (though a good feature idea for future development).

jluethi:
Awesome, looks great like this.
Does it still work for users that don't have custom tasks (i.e. they'd just see admin first)?

tcompa:
There are still use cases which are not covered (see also https://github.com/fractal-analytics-platform/fractal-web/pull/233#issuecomment-1619641414 by @jluethi).

I create the following tasks
```
name="test", owner="admin", source="admin:a", version=null
name="test", owner="admin", source="admin:b", version=null
name="test", owner="user", source="user:a", version=null
```
and they are correctly displayed as
```
admin
    test
user
   test
```
Then I click on the first `test` task name (i.e. the one belonging to `admin`), and I'm shown
```
 Select task version 
    Not specified
    Not specified
    Not specified
```

We are hitting two issues here:
1. If we are separating admin/test and user/test, then "Select task version" should only show two items (the ones with owner=admin) rather than three.
2. If the version is not there, this selection is not meaningful (independently on point 1). I think that "Not specified" should always be replaced by the task source, and "Select task version" should be "Select task version/source".

jluethi:
Agreed, these two things are still missing.

> 1. If we are separating admin/test and user/test, then "Select task version" should only show two items (the ones with owner=admin) rather than three.
> 2. If the version is not there, this selection is not meaningful (independently on point 1). I think that "Not specified" should always be replaced by the task source, and "Select task version" should be "Select task version/source".

In addition:

> Search is not functional yet, right? But great idea, would love search when adding the tasks!

I'd see these 3 areas as the last things remaining here

rkpasia:
@tcompa I noticed this bug

> If we are separating admin/test and user/test, then "Select task version" should only show two items (the ones with owner=admin) rather than three.

And made a fix in [5a44460](https://github.com/fractal-analytics-platform/fractal-web/pull/233/commits/5a44460e8ea7b36f2c8b17775c128f68aefd4eb3).

Now it should work.

Related to

> If the version is not there, this selection is not meaningful (independently on point 1). I think that "Not specified" should always be replaced by the task source, and "Select task version" should be "Select task version/source".

Now if the version is not specified, the `source` property is shown. Moreover, if there are multiple tasks with the same version, only `source` should be displayed in the task version/source selection.

tcompa:
> Moreover, if there are multiple tasks with the same version, only source should be displayed in the task version/source selection.

Is this already implemented?

I have tasks with versions "1", "2", and "2", and in the drop-down menu I see
* v1
* v2
* user:3 (which is the source of the third task)

I don't have a strong opinion here, but I would be OK with your idea (if there are multiple tasks with the same version, only source should be displayed in the task version/source selection) - once it's implemented.

rkpasia:
> If the version is not there, this selection is not meaningful (independently on point 1). I think that "Not specified" should always be replaced by the task source, and "Select task version" should be "Select task version/source".

On this we could go even further, given:
- A user want to select a task to insert in the workflow
- Opens the new selection tab
- Selects the `Users tasks` tab
- Now, if a user selects a task that has multiple **same** versions in the server, but they are assigned different owners, in the select version/source dropdown the user should be shown just the `version` name and not the `source` name.

Currently, the client, if multiple **same** task versions are present, they are replaced with `source` name instead of `version`, by default.
Should we discriminate by user?

jluethi:
> if there are multiple tasks with the same version, only source should be displayed in the task version/source selection

I'd agree with that as well

jluethi:
> Now, if a user selects a task that has multiple same versions in the server, but they are assigned different owners, in the select version/source dropdown the user should be shown just the version name and not the source name.
> Currently, the client, if multiple same task versions are present, they are replaced with source name instead of version, by default.
> Should we discriminate by user?

If we easily can, then yes.

rkpasia:
Ok, so a quick recap:

- Task can be picked by two groups, `common` and `user`
- Common tasks are again grouped by `package`
  - When selecting a task that belongs to a package, if there are multiple tasks with same name and version, `source` property is used to select the task instead of `version`.
- Users tasks are grouped by `owner`
  - When selecting a task that belongs to a user, if there are multiple tasks with same name and version, `source` is used instead of `version`. (Same as _common_ group)
- The select box groups tasks automatically through headers
- We should include the search feature to improve further the selection experience once there will be many tasks available.

jluethi:
> - Task can be picked by two groups, common and user
> - Common tasks are again grouped by package
> - When selecting a task that belongs to a package, if there are multiple tasks with same name and version, source property is used to select the task instead of version.
> - Users tasks are grouped by owner
> - When selecting a task that belongs to a user, if there are multiple tasks with same name and version, source is used instead of version. (Same as common group)
> - The select box groups tasks automatically through headers
> - We should include the search feature to improve further the selection experience once there will be many tasks available.

Agreed with all of this, this is the behavior we want. Great summary!

>  then "Select task version" should only show two items (the ones with owner=admin) rather than three

i.e. only the version or source for the task belonging to a given user should be shown in the dropdown, not all all available versions for this task among the users.

tcompa:
> When selecting a task that belongs to a package, if there are multiple tasks with same name and version, source property is used to select the task instead of version.

I suggested this feature for cases where we do include additional requirements for task collection (e.g. the python version, or some additional extras). This could generate tasks that have the same (package, name, version), and then we should use the source to identify them.
Note that implementing a full parsing of the source (that also reads the python version and/or the extras) doesn't seem very future-proof. Also, I don't expect that explicitly setting the python version is something we are going to need often.

jluethi:
> I suggested this feature for cases where we do include additional requirements for task collection (e.g. the python version, or some additional extras). This could generate tasks that have the same (package, name, version), and then we should use the source to identify them.

This sounds future-proof enough to me: 
In most cases, name + version are unique, thus show them. 
In the special cases where they are not, show the source (not as pretty, but guaranteed to be unique)

rkpasia:
> We should include the search feature to improve further the selection experience once there will be many tasks available.

Has a related problem #238

PR #234, opened 2023-06-29
Introduce unsaved arguments changes in workflow page

rkpasia:


tcompa:
This looks to me like the expected behavior, thanks!
For the record: an informative popup is displayed when:
* trying to run the workflow with unsaved changes  in the current workflow task
* trying to switch to a different workflow task with unsaved changes in the current one

We should then check whether it is doable/OK to do the same to also to prevent tab closing (ref https://svelte.dev/repl/a95db12c1b46433baac2817a0963dc93?version=4.0.1). Likely we won't be able to display a custom message (see e.g. [here](https://bugs.chromium.org/p/chromium/issues/detail?id=587940)), but we should at least be able to trigger a generic one (in Firefox it reads "This page is asking you to confirm that you want to leave — information you’ve entered may not be saved.").

tcompa:
Current status:

Screenshot from 2023-06-29 15-50-33

jluethi:
And while looking at that screenshot: Can we place save & discard in the same arrangement? i.e. now, at the bottom of the page, the save button is on the left, but in the dialog it's on the right. It'd prefer it to be on the right in both cases (unless someone has strong opinions here)

rkpasia:
I'm not sure about this:

> If you change something, try to switch to a different task, then say discard, you actually switch to that task. If you say save, you remain on the task you were on (as if the task selection change gets ignored).

This I think would really be a decision about user experience and what an user would expect to happen.

## TLDR;
I think that for newer users preventing to change the active workflow task would be a better experience. I would sacrifice the needs of "power" users in this situation.

---
More context, for instance:

1. As a user I made some changes to the args.
2. Say I don't see the 'Save changes' button and just move on with my business.
3. I get prompted a modal that says: would you like to save changes?
4. I say yes and move on.

After 4, what would I think?
1. Oh I need to save! Wait, what am I saving? Is it right? let me double check. (newer user)
2. Oh I forgot to save, yeah save and move on! (more advanced user)

In (1) as a first user, it could be mislead by the change of active workflow task, also because the 'changes have been saved' label would be displayed showing different arguments.
What could we expect the user would think in that situation? _Oh, every argument is different than the ones I set._
Then suddently, the user would realize for sure that the changes have been saved but the UI has already moved on.
This experience could feel a little bit rough for the guts of the newer users.

It's a personal opinion anyway, we could go with whatever 👍🏼

jluethi:
Thanks @rkpasia , those are good points. In that case, let's stay with the behavior of not switching tasks when the user clicks save.

Does the new catch for unsaved changes also apply if the user e.g. clicks on the task page in the header? Or just when the tab is closed?

rkpasia:
[acc307a](https://github.com/fractal-analytics-platform/fractal-web/pull/234/commits/acc307a633ca1dab3745061fd3f47745bc12609f) Should solve any of the following:

- Prevent an user from closing the window if there are unsaved changes
- Prevent an user from navigating to another page
- Prevent an user to trigger a workflow
- Prevent an user to switch active workflow task

I think that this should check all the situations we would prevent an user to go away if there are unsaved changes.

tcompa:
Testing with acc307a633ca1dab3745061fd3f47745bc12609f

> Prevent an user from navigating to another page

I confirm it works as expected

> Prevent an user to trigger a workflow

I confirm it works as expected

> Prevent an user to switch active workflow task

I confirm it works as expected

> Prevent an user from closing the window if there are unsaved changes

There is a glitch, since two popups appear (the generic browser one and ours). Not a huge problem, but it's just FYI

Screenshot from 2023-06-30 12-48-05

rkpasia:
> There is a glitch, since two popups appear (the generic browser one and ours). Not a huge problem, but it's just FYI

The browser-specific alert is triggered by the browser itself and depends on the specific policies applied.

In certain browser the application lifecycle is stopped until the user takes an action on the browser alert. In others, as you cited, the lifecycle moves on and the background modal is shown.

This is something that I'm not sure we can control.

tcompa:
> The browser-specific alert is triggered by the browser itself and depends on the specific policies applied.
> 
> In certain browser the application lifecycle is stopped until the user takes an action on the browser alert. In others, as you cited, the lifecycle moves on and the background modal is shown.
> 
>  This is something that I'm not sure we can control.

Thanks for the clarification!

Given what you say, the current behavior looks OK to me (it would be nicer to only have one, but I don't think our effort should be in optimizing the closing-tab-or-window-with-unsaved-changes behavior).

tcompa:
@jluethi the latest version in this branch looks good to me, if you can double-check that it matches with your expectations we can merge this one

jluethi (approved):
Also just tested it, this is great! It stops the users correctly from leaving the page in all the scenarios that I can think of! Ready to merge from my perspective.

jluethi:
@tcompa Just done :) Agreed, this is ready to be merged

tcompa (approved):

jluethi:
We do see this warning in the review, is this relevant?

Screenshot 2023-06-30 at 13 38 21

rkpasia:
It's a dependency that is loaded through network via a CDN.
The warning is related to the missing reference to a local package that is not loaded but will be present, say, at runtime.

rkpasia:
This won't prevent che cliente from running correctly.

jluethi:
Perfect! It's running fine on my side, just briefly wanted to check :)

PR #237, opened 2023-07-03
Set up preliminary docs structure

tcompa:


rkpasia:
Would be nice to have documented how the authentication works.

tcompa:
Thanks a lot @rkpasia for what you could fill in, especially in `structure.md`. I'm merging this PR for now, so that we have a starting point for future improvements.

Issue #265, opened 2023-08-25
Switch from using `formData` to binding variables


tcompa:
This was added e.g. in the task-edit as part of #262. This is the preferred approach, and it should be used more homogeneously in all forms.

tcompa:
Is think this was mostly covered with #288 (@zonia3000 let me know if I'm wrong), as I can only see `formData` in
```console
$ git grep formData src/

src/routes/auth/login/+page.server.js:          const formData = await request.formData();
src/routes/auth/login/+page.server.js:                  authData = await userAuthentication(fetch, formData);
```

We may want to also update that file, but perhaps as part of a broader user/login review (see e.g #246, #4 and #277).

zonia3000:
Yes, `formData` is now used only in the login calls. I would refactor them when we will introduce OAuth2.

tcompa:
Great. Then I'm closing this one in favor of #246.

PR #288, opened 2023-09-19
Refactoring of error propagation and more

### LINKED ISSUE #178: Replace "The server was not able to stop the job" with actual error message
If a cancel-a-job operation fails, the user should see a message that includes the original fractal-server error message.
Currently the user only sees "The server was not able to stop the job".

### LINKED ISSUE #248: To check: Deleting a workflow in a project that has been run before fails with a server 500 error
We observed this _live_ during last week discussions. To be verified.

### LINKED ISSUE #250: Correctly propagate fractal-server error (upon deleting a dataset that has been run in a job)
When a user tries to delete a dataset which has already been used in a job, this fails server-side. See here for discussion of it: https://github.com/fractal-analytics-platform/fractal-server/issues/667

This is not displayed nicely in the web interface though. The actual interface just doesn't do anything, but also doesn't show the error. In the console, the following error shows up:
```
start.d697be3d.js:1     DELETE https://fractal.mls.uzh.ch/projects/1/datasets/2 500 (Internal Server Error)
window.fetch @ start.d697be3d.js:1
h @ projects-_projectId_-page.svelte.73b41b8d.js:1
y @ ConfirmActionButton.8aeb571e.js:1
projects-_projectId_-page.svelte.73b41b8d.js:1 Error while deleting dataset: The dataset was not deleted by the server
```

We'd want the user to see this error (or a nicer form of it) in the interface until it's actually possible to delete those datasets.

### LINKED ISSUE #289: Cannot create a new task without setting version
version 0.5.6

Error message
```
There has been an error, reason:

{
  "detail": [
    {
      "loc": [
        "body",
        "version"
      ],
      "msg": "String attribute 'version' cannot be empty",
      "type": "value_error"
    }
  ]
}
```

Request payload
```
-----------------------------401338746928919240913247038010
Content-Disposition: form-data; name="name"

c
-----------------------------401338746928919240913247038010
Content-Disposition: form-data; name="command"

c
-----------------------------401338746928919240913247038010
Content-Disposition: form-data; name="source"

c
-----------------------------401338746928919240913247038010
Content-Disposition: form-data; name="version"


-----------------------------401338746928919240913247038010
Content-Disposition: form-data; name="input_type"

c
-----------------------------401338746928919240913247038010
Content-Disposition: form-data; name="output_type"

c
-----------------------------401338746928919240913247038010--
```

Fastapi logs:
```
INFO:     127.0.0.1:47682 - "POST /api/v1/task/ HTTP/1.1" 422 Unprocessable Entity
```

### LINKED ISSUE #291: Review use of browser cache for task-collection events
To reproduce:
1. Trigger a task collection
2. Remove the task collection from the DB (e.g. because you are starting a fresh fractal-server+DB instance)
3. Browse to http://localhost:5173/tasks

Because the list of old task-collections is stored in the browser cache, an API call is sent out to fractal-server to retrieve the corresponding item.
This one fails in fractal-server
```
INFO:     127.0.0.1:46872 - "GET /api/v1/task/collect/1?verbose=True HTTP/1.1" 404 Not Found
```
and then we hit this error:
image

I think these calls (the ones fetching the old task collection) should be more "forgiving". If the corresponding GET fails, we can e.g. modify the entry in the list, or plain remove it, but this error is not really useful.

zonia3000:
## Checklist before merging
- [x] Let's also add a few words in some docs file (maybe https://github.com/fractal-analytics-platform/fractal-web/blob/main/docs/structure.md)

---

I think that when an error happens while using the modal handled by the `ConfirmActionButton` component, the error should be displayed inside the modal and the modal should not be closed in that case, so I modified the `handleCallbackAction()` function inside the `ConfirmActionButton` component to catch a specific `AlertError` that can be thrown by the custom callback set on the component. On all the other cases (when we are not using the modal) I expect to check the `response.ok` value and open directly a `StandardErrorAlert`, without any kind of error rethrowing logic.

fractal-delete-dataset-error

I introduced the `src/routes/api/[...path]/+server.js` file, that acts like a proxy for the Python backend. The following features have been rewritten calling these proxy endpoints and removing the duplication of calls:

* creation and listing of projects
* CRUD of datasets
* CRUD of dataset resources

Note: I'm adding the `@type` comments in various places. This suggests the types to the IDE and makes it happier without having to move to TypeScript.

Closes #250
Closes #289 
Closes #248
Closes #178
Closes #291


tcompa:
Here is a first round of (human) tests

# creation and listing of projects

* creating a new project -> OK
* creating a new project with the same name as another one -> error is displayed correctly -> OK
* listing projects -> all as expected (both after success/failure in project creation) -> OK

# CRUD of datasets
* read a dataset -> OK
* edit dataset name -> OK
* edit dataset read-only property -> OK
* add dataset type -> OK
* unset dataset type: this has a glitch unrelated to this PR (as already identified by @zonia3000), since the action goes through but with no effect (likely because of stripping `null`s from request bodies)
* deleting a dataset -> OK
* deleting a dataset related to a job -> error is displayed correctly -> OK


# CRUD of dataset resources
* creating resource -> OK
* creating resource with non-absolute path -> error is displayed correctly -> OK
* deleting resource -> OK
* viewing existing resources -> OK

zonia3000:
Closes #289

tcompa:
> Closes #289

I confirm that https://github.com/fractal-analytics-platform/fractal-web/pull/288/commits/9fcf784f69e4408f5749da0d8d7f10367301a560 fixes #289 (i.e. I can now create a task without setting its `version` in the form).

tcompa:
Some more tests:

- [x] Successful task collection
- [x] Failed task collection -> error displayed
- [x] Successful task edit
- [ ] Failed task edit -> to check

tcompa:
- [x] Failed workflow import -> error displayed
- [x] Successful workflow import 

GLITCH: after a failed workflow import, close the modal, then perform a successful import. The red alert of the first error is still there.

zonia3000:
> GLITCH: after a failed workflow import, close the modal, then perform a successful import. The red alert of the first error is still there.

This should be ok now

zonia3000:
> unset dataset type: this has a glitch unrelated to this PR, since the action goes through but with no effect (likely because of stripping nulls from request bodies)

I tried to set type to null, but I receive `String attribute 'type' cannot be None`, so it seems that it can't be unset in any case.

tcompa:
#### [Refactored import workflows, project deletion and project name update](https://github.com/fractal-analytics-platform/fractal-web/pull/288/commits/96df0ceaddddddbf7312b94fe92ac44ea4b825ff)

- [x] successful workflow import
- [x] failed workflow import
- [x] successful project-name update
- [x] failed project-name update (see below)
- [x] successful project delete

A failed project-name update results into a very compact error message, with no details:
Screenshot from 2023-09-20 14-53-16
 
I guess this was on purpose? Fine with me, when such a limited-scope action fails, but in general let's try to include the error message.

#### [Fixed workflow import error glitch](https://github.com/fractal-analytics-platform/fractal-web/pull/288/commits/b0775c899f4e372f83f0c084c3569c9112d27336)

Fixed, thanks!

tcompa:
> > unset dataset type: this has a glitch unrelated to this PR, since the action goes through but with no effect (likely because of stripping nulls from request bodies)
> 
> I tried to set type to null, but I receive `String attribute 'type' cannot be None`, so it seems that it can't be unset in any case.

It's OK.
This would fit into a broader review of request body validation, but it's out of scope here.

zonia3000:
I'm seeing that the modals displaying the job info and logs are calling the API 3 times because the HTTP call is done inside the `afterUpdate()` function, that is invoked multiple times during the rendering. Tomorrow I would like to fix also this. In any case there are just a few endpoints to be changed, so I should be able to conclude the refactoring.

tcompa:
Since this PR (correctly!) streamlines the propagation of fractal-server errors, we now also observe an error message that up to now was hidden:
* https://github.com/fractal-analytics-platform/fractal-web/issues/291

It's not a major issue (as it only appears in specific dev use cases), but we should eventually review the way we clean that list from the cache (not necessarily as part of this PR).

tcompa:
Testing the latest commits:

## Refactored create workflow task

It works successfully.
I have no way to make it fail.

## Refactored delete workflow task

It works successfully.
I have no way to make it fail.

## Refactored apply workflow

Failure (Input dataset has empty resource_list) -> ok
Failure (Incompatible types..) -> ok
Successful call -> ok

## Refactored export workflow

It seems to work as expected

## Refactored get available tasks

I see the correct lists, both for common and users' tasks.

## Refactored stop job

The local-backend error ("detail": "Stopping a job execution is not implemented for FRACTAL_RUNNER_BACKEND=local.") is displayed correctly -> fixes #178

tcompa:
Things that we can quickly review together (and possibly postpone to different PRs):
* https://github.com/fractal-analytics-platform/fractal-web/pull/288#issuecomment-1728088855 (repeated API calls for job logs/info)
* https://github.com/fractal-analytics-platform/fractal-web/pull/288#issuecomment-1728967924 (#291)
* Decision about "short" error messages, see https://github.com/fractal-analytics-platform/fractal-web/pull/288#issuecomment-1727671135

zonia3000:
> * [Refactoring of error propagation and more #288 (comment)](https://github.com/fractal-analytics-platform/fractal-web/pull/288#issuecomment-1728088855) (repeated API calls for job logs/info)
> * [Refactoring of error propagation and more #288 (comment)](https://github.com/fractal-analytics-platform/fractal-web/pull/288#issuecomment-1728967924) ([Review use of browser cache for task-collection events #291](https://github.com/fractal-analytics-platform/fractal-web/issues/291))

fixed

tcompa:
Thanks for aa5542d, I confirm it works as expected (or at least as _I_ expected):

image

tcompa:
I'm not sure that 5fecc0ff043f1503335af681b9143866d61feb30 works as expected, with respect to #291.
* I perform some task collections (either successful or failed).
* I go and remove one of them from the `state` table in the db
* I refresh the page, and they are all still listed in the "Task collections" list
* I click on "info" for the one I removed, and then I get something as in the screenshot below. Meanwhile, fastapi responded with a 404, as in
```
INFO:     127.0.0.1:46274 - "GET /api/v1/task/collect/1/?verbose=True HTTP/1.1" 307 Temporary Redirect
INFO:     127.0.0.1:46270 - "GET /api/v1/task/collect/1?verbose=True HTTP/1.1" 404 Not Found
```

Then in which cases do we automatically remove an item from the list?




image

tcompa:
Related to the last comment, note that the logs shown in the screenshot (below the error message) most likely belong to a different item, and not to the one that is being fetched now (which is not surprising, since that fetch will fail with 404).

tcompa:
Another minor glitch question:
I do observe logs like
```
INFO:     127.0.0.1:46274 - "GET /api/v1/task/collect/1/?verbose=True HTTP/1.1" 307 Temporary Redirect
INFO:     127.0.0.1:46270 - "GET /api/v1/task/collect/1?verbose=True HTTP/1.1" 404 Not Found
```

This is probably due to the trailing slash, right? In [fractal-server](https://fractal-analytics-platform.github.io/fractal-server/openapi/) we do not add a trailing slash when the path ends with an ID (while we are a bit less consistent when the path ends with a word, because then it depends on whether the endpoint is our own or one that we take from fastapi-users).

zonia3000:
> Also remove/refactor src/routes/projects/[projectId]/jobs/[jobId]/logs/+server.js

[43080bb](https://github.com/fractal-analytics-platform/fractal-web/pull/288/commits/43080bb221eb958ba25e69c9450266096901ff1c) refactored the download of job logs. Notice that while passing through the "Svelte proxy" the response body isn't streamed, so the user can experience a sort of lag during the download if the content of the file is very big. In any case this happened also before, because the response was read and then converted to a blob object before the download, so I assume it is not a problem. This note is just a consideration in case we will need to handle big files in the future.

zonia3000:
> * Review the status of [Switch from using `formData` to binding variables #265](https://github.com/fractal-analytics-platform/fractal-web/issues/265)

The only remaining action that uses formData is the login. I wouldn't touch this part in this PR, since it involves the call of `/auth/` API and we should add another set of "proxy endpoints" to handle it.

tcompa:
> I'm not sure that [5fecc0f](https://github.com/fractal-analytics-platform/fractal-web/commit/5fecc0ff043f1503335af681b9143866d61feb30) works as expected, with respect to #291.
> 
>     * I perform some task collections (either successful or failed).
> 
>     * I go and remove one of them from the `state` table in the db
> 
>     * I refresh the page, and they are all still listed in the "Task collections" list
> 
>     * I click on "info" for the one I removed, and then I get something as in the screenshot below. Meanwhile, fastapi responded with a 404, as in
> 
> 
> ```
> INFO:     127.0.0.1:46274 - "GET /api/v1/task/collect/1/?verbose=True HTTP/1.1" 307 Temporary Redirect
> INFO:     127.0.0.1:46270 - "GET /api/v1/task/collect/1?verbose=True HTTP/1.1" 404 Not Found
> ```
> 
> Then in which cases do we automatically remove an item from the list?
> 
> image

I confirm that after [32a6d3c](https://github.com/fractal-analytics-platform/fractal-web/pull/288/commits/32a6d3cf4d9256e8c5402ad4fc9433d6aa0887ce) I observe the expected behavior.

tcompa:
> > Also remove/refactor src/routes/projects/[projectId]/jobs/[jobId]/logs/+server.js
> 
> [43080bb](https://github.com/fractal-analytics-platform/fractal-web/pull/288/commits/43080bb221eb958ba25e69c9450266096901ff1c) refactored the download of job logs. Notice that while passing through the "Svelte proxy" the response body isn't streamed, so the user can experience a sort of lag during the download if the content of the file is very big. In any case this happened also before, because the response was read and then converted to a blob object before the download, so I assume it is not a problem. This note is just a consideration in case we will need to handle big files in the future. It could be solved setting up a "real" proxy between the front-end and the Python API.

Thanks for the note. Agreed, this is not a priority now.

Meanwhile I tested the download-logs feature still works as expected.

tcompa:
> I think the following one is a bug:
> 
> If I submit a workflow, and if I do set some specific values of the first/last task to run, the payload for the `apply` endpoint looks like
> 
> ```
> {
>         'worker_init': None,
>         'first_task_index': None,
>         'last_task_index': None,
>     }
> ```
> 
> while I would expect `first_task_index` and `last_task_index` to be something (e.g. 0 and 2).

We found the error source with Sonia (a typo in the parameter name), and it's easy to fix.

Other than that, I don't have anything more to add for the moment.

Issue #108, opened 2023-04-18
"Add-resource" dialogue pops up again if manually closed after adding a resource


jluethi:
Currently, when a user adds a new resource, a confirmation is shown in the add resource pop-up. The pop-up then closes after about 2 seconds.
See intended behavior here: https://drive.google.com/file/d/1s9W-aJRlcbl9RtrElWCBYwMTZTOdnVxV/view?usp=sharing

If the user is impatient and closes the resource adding modal before it closes itself, a new & empty add resource modal pops up, which is quite confusing. See example here: https://drive.google.com/file/d/1OFV0NS785sIC5tr6KbeFWJo2U0eQxfJI/view?usp=sharing

tcompa:
> If the user is impatient and closes the resource adding modal before it closes itself, a new & empty add resource modal pops up, which is quite confusing

Agreed: I also often hit this issue, and find it confusing.

jluethi:
Glad to hear I'm not the only impatient user ;) 

I don't mind the behavior of the modal not closing immediately, but closing after 2s (=> gives users a chance to see the confirmation). But I should be able to close it faster without it popping back up 😅

Alternatively, another valid behavior would be to close when the user clicks Create and show the confirmation that it has been added somewhere else. Or drop that confirmation, because it does show up in the list of resources anyway, so not really necessary to have a separate confirmation.

rkpasia:
We could go with a different kind of confirmation that displays within the page instead of within the context of the modal.

jluethi:
> We could go with a different kind of confirmation that displays within the page instead of within the context of the modal.

That sounds preferable :)

tcompa:
Closing as obsolete, see e.g. https://github.com/fractal-analytics-platform/fractal-web/pull/306#issuecomment-1748482812.

PR #306, opened 2023-10-05
Modal refactoring

### LINKED ISSUE #264: Implement a custom event for modal dismiss
Sometimes it is necessary to execute custom logic on modal close (see #262). Currently we have associated this custom logic to the close button, but a modal can be closed also clicking on the X button on the top right corner or clicking outside the modal, so at the moment a [static backdrop](https://getbootstrap.com/docs/5.3/components/modal/#static-backdrop) has been set to force the user to close the modal only by clicking on the button. Since this can be annoying it should be better to define a custom Svelte event to execute this custom logic for each closing method.

### LINKED ISSUE #307: Review `src/lib/types.d.ts`
I don't fully grasp the way this file is used, but it looks like it differs from the actual type of a task - see e.g. the `TaskRead` entry in https://fractal-analytics-platform.github.io/fractal-server/openapi.json.

zonia3000:
## Checklist before merging

I introduced a new generic `Modal` component on which the developer can attach functions to the Bootstrap modal events, so that we can detect the modal closure regardless of the chosen mode (button, esc key, click on backdrop). It also provides generic methods for displaying an error inside the modal and automatically cleaning up the previous errors if the modal is closed and then opened again.

After this refactoring the delete workflow confirm modal behaves like delete dataset confirm modal: the error is always displayed inside the modal and the modal is not closed when an error happens.

Moreover, it was now possible to remove the static backdrop mentioned in #264, meaning that now the edit task modal can be closed in all the possible ways.

I've also investigated the issue of the backdrop remaining stuck when clicking on the back browser button. It seems to be related to Svelte invalidation not being triggered for the modal when going to the previous page and I fixed it adding a check directly in the layout page.

All the modals involved in the refactoring:

* generic confirm action
* edit task (it is now possible to close the modal in all the possible ways)
* import workflow
* create dataset resource
* update datatset
* job info
* job logs
* project info
* task collection logs
* task info
* edit project
* dataset meta properties
* unsaved changes
* new workflow task
* edit workflow properties
* change workflow task order
* run workflow

Closes #264
Closes #307

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -1,0 +1,33 @@
+name: End-to-end tests
+
+on:
+  push:
+    branches: ['main']
+  pull_request:
+    branches: ['main']
+
+jobs:
+  end_to_end_tests:
+    name: 'Node ${{ matrix.node-version }}'
+    runs-on: ubuntu-20.04
+    timeout-minutes: 10
+
+    strategy:
+      matrix:
+        node-version: ['16', '18', '20']
+
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v4
+
+      - name: Set up node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run playwright tests (not implemented)
+        run: echo "Not implemented"
--- a/.github/workflows/lint_and_build.yaml
+++ b/.github/workflows/lint_and_build.yaml
@@ -1,4 +1,4 @@
-name: Check&Build
+name: Lint and build
 
 on:
   push:
@@ -7,18 +7,18 @@ on:
     branches: ['main']
 
 jobs:
-  check_and_build:
+  lint_and_build:
     name: 'Node ${{ matrix.node-version }}'
     runs-on: ubuntu-20.04
     timeout-minutes: 10
 
     strategy:
       matrix:
-        node-version: ['16', '18']
+        node-version: ['16', '18', '20']
 
     steps:
       - name: Check out repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up node
         uses: actions/setup-node@v3
@@ -31,9 +31,6 @@ jobs:
 
       - name: Run static code analysis
         run: npx eslint .
-
-      - name: Run vitest tests
-        run: npm run test
 
       - name: Build site
         env:
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -1,0 +1,33 @@
+name: Unit tests
+
+on:
+  push:
+    branches: ['main']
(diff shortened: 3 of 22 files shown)


tcompa:
> After this refactoring the delete workflow confirm modal behaves like delete dataset confirm modal: the error is always displayed inside the modal and the modal is not closed when an error happens

I agree with this choice, and here are some examples that match with the expected behavior:
Screenshot from 2023-10-05 11-18-31
Screenshot from 2023-10-05 11-18-00
Screenshot from 2023-10-05 11-16-35

Perhaps in some cases we may make the modal a bit wider, since error messages tend to be verbose, but this can also be postponed to future "style" cleanups

tcompa:
> I've also investigated the issue of the backdrop remaining stuck when clicking on the back browser button. It seems to be related to Svelte invalidation not being triggered for the modal when going to the previous page and I fixed it adding a check directly in the layout page.

In my first tests this generally works, and I observe this behavior:
* I am on page A
* I go to page B
* I open a modal
* I click backspace
* I am back to page A

tcompa:
> edit task (it is now possible to close the modal in all the possible ways)

I tried by:
* "close" button
* clicking outside the modal
* backspace

In all three cases, any change that I made inside the modal before closing it (e.g. changing `type` from `zarr` to `somethingelse`) were not present upon re-opening it.

This matches with expectations.

tcompa:
> It also provides generic methods for displaying an error inside the modal and automatically cleaning up the previous errors if the modal is closed and then opened again.

I confirmed this works, for the add-resource-to-dataset modal.
Here is the error:
image
And this is after I close the modal, and re-open it:
image

tcompa:
A question: does the generic `Modal` component include any instruction on when to close the modal?

For instance when a resource is successfully added to a dataset we see the following, and we need to click somewhere to close the modal
image

In this case, would it make sense to close it automatically?

---

Note: this specific modal was part of an old issue (https://github.com/fractal-analytics-platform/fractal-web/issues/108), which probably was already somehow fixed in the past. From the old issue:

> Currently, when a user adds a new resource, a confirmation is shown in the add resource pop-up. The pop-up then closes after about 2 seconds.

This is not the case any more.

> If the user is impatient and closes the resource adding modal before it closes itself, a new & empty add resource modal pops up, which is quite confusing

This is also not happening any more.

Therefore I am now closing that old issue, and the only question left is whether it makes sense to automatically close the modal after it "success".

zonia3000:
> In general I think that the top-right X button is useful and doesn't introduce too much "noise", and I would always include it in all modals. No problem if there are different opinions here ;)

I've included the top-right X button in the generic `Modal` component.

> A question: does the generic `Modal` component include any instruction on when to close the modal?

The generic modal provides a method to automatically close the modal on a successful action, but one can choose to use it or not. I've refactored the following modals to close automatically on success:

* Create dataset resource
* Edit project
* Import workflow (this now redirects to the newly created workflow, as happens for the simple workflow creation)
* Edit project properties
* New workflow task
* Edit Workflow properties
* Update dataset properties

After the closure, a success alert message is displayed in the page for a few seconds.

Issue #309, opened 2023-10-09
Add "Show history" to dataset page


tcompa:
Right now we have this logic in `src/routes/projects/[projectId]/datasets/[datasetId]/+page.svelte`:
```svelte
		{#if dataset && Object.keys(dataset.meta).length > 0}
			<button class="btn btn-light" data-bs-target="#datasetMetaModal" data-bs-toggle="modal">
				<i class="bi-arrow-up-right-square" />
				Show meta properties
			</button>
		{/if}
```

1. We should add a similar logic for the `dataset.history` property, as soon as this is part of a fractal-server release (issue: https://github.com/fractal-analytics-platform/fractal-server/issues/838).
2. In my view, we should not make these buttons (the ones for meta and history) disappear when the corresponding property is missing. I think they should always be present, since this provides a more consistent experience of the Dataset page. We can find simple ways of handling a null/empty property: we either disable the button, or we open the modal and say "this property is missing/empty/...".

tcompa:
> 2\. In my view, we should not make these buttons (the ones for meta and history) disappear when the corresponding property is missing.

Fixed with https://github.com/fractal-analytics-platform/fractal-web/pull/310/commits/6cc0409b9e4cd864c1e02bad0196b389e339eec5.

PR #316, opened 2023-10-13
Add show-history button in dataset page (ref #309)

tcompa:
I'm adding this feature in an experimental way, as it's useful for testing the upcoming (unreleased, as of now) fractal-server version, based on https://github.com/fractal-analytics-platform/fractal-server/pull/898.

Here are some screenshots of how it would look:
image
image

I think we will need to slightly decrease the amount of information shown in there.
The simplest way would be to hide the `args_schema` property (or replace it with the string "HIDDEN", or something like that). Each history item has a required `workflowtask` property, then a required `task` property, then an optional `args_schema` property. If `item.workflowtask.task.args_schema` exists, we could modify it.

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Unreleased
 
+* Added dataset history modal (\#316).
 * Fixed bug with save/discard changes in workflow task (\#315).
 * Fixed first-run redirects issue (\#315).
 
--- a/src/lib/types.d.ts
+++ b/src/lib/types.d.ts
@@ -36,9 +36,9 @@ export type Dataset = {
   name: string
   type?: string
   meta: object
-  history: object
+  history: Array<object>
   read_only: boolean
   id: number
-  resource_list: Resource[]
+  resource_list: Array<Resource>
   project_id: number
 }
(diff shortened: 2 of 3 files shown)


tcompa:
This can be tested with fractal-server 1.3.12a0

tcompa:
There's a minor glitch we may still want to fix.
When using fractal-server <1.3.12a0, the GET-dataset response does not have a property history. This is triggering some errors, since in the modal we try to convert an `undefined` into an `object`, here:
```svelte
		<svelte:fragment slot="body">
			<ul class="list-group">
				{#if dataset.history && Object.keys(dataset.history).length > 0}
```

I fixed it with [ad3bf42](https://github.com/fractal-analytics-platform/fractal-web/pull/316/commits/ad3bf42e517929a6e2bc35bab0235274cfae789c), so that fractal-web is a bit more robust when working with a slightly older fractal-server instance.

@zonia3000 is this change OK for you?

Also: is there anything else missing here? Otherwise I'd say we can merge.

zonia3000:
> @zonia3000 is this change OK for you?
> 
> Also: is there anything else missing here? Otherwise I'd say we can merge.

Fine for me. I think we can merge.

Issue #322, opened 2023-10-23
Heavy performance hit when many items in a dictionary of a workflow task


jluethi:
I just observed that the performance of workflow task editing gets very bad when there are (reasonably) many entries for allowed channels in the Create OME-Zarr (multiplexing) task.

Specifically, the user has 6 cycles (6 keys in the dictionary for allowed channels) and each entry has 4 channel objects. Now, just loading the workflow page makes the browser unresponsive, it can take 5-10s (sometimes with a "browser unresponsive" showing up).

Let's evaluate the performance here and see if this can be reproduced. I see the same by logging in as that user on the UZH deployment.
Here is the example workflow
[workflow-export-4i per well workflow-1698072006544.json](https://github.com/fractal-analytics-platform/fractal-web/files/13072234/workflow-export-4i.per.well.workflow-1698072006544.json)

tcompa:
A couple of sanity checks (cc @zonia3000):

1. After the new playwright PR, are we now running coverage tools also in non-dev environments (e.g. when we do something like `npm run build; node build`)?
2. If so, can we disable it? (unless it's known that it has no impact on performance)

I cannot say whether this is related to the current issue, but I'm asking because the current `npm run dev` logs includes lines like
```
vite:istanbul> Sourcemaps was automatically enabled for code coverage to be accurate.
 To hide this message set build.sourcemap to true, 'inline' or 'hidden'.
```
and I was wondering if this has implications re: performance.

jluethi:
Ok, that's also interesting to check. The performance issue I observed was specific to the one workflow task page with the heavy load, the other workflow task pages were fine

tcompa:
Another for-the-record remark (related, but not necessarily relevant for the current issue):

I think we should proceed and implement https://github.com/fractal-analytics-platform/fractal-web/issues/161: instead of only using some JSON Schema examples, we should (also) use the actual ones from fractal-tasks-core.

Once this is in-place, we can also add a benchmark of a couple of "complex" situations (e.g. the create-ome-zarr-multiplexing JSON Schema, with a somewhat large object of arguments to validate). For the moment I'm rather thinking about by-hand benchmarks (measure the time a certain set of operations take, and then check that it's below some threshold) and not about formal benchmarks (_à la_ https://github.com/tinylibs/tinybench); but we can of course re-discuss it.

zonia3000:
> 
> ```
[...]

PR #324, opened 2023-10-24
Fixed argument popovers performance issue

zonia3000:
## Checklist before merging

I've run the Firefox profiler using the workflow file provided by @jluethi in #322.

It detects that the performance issue is related to the file `PropertyDescription.svelte`:

Screenshot from 2023-10-24 08-50-18

That component displays the description of each argument inside a popover. However it was initializing every time all the popovers of the page for each component, resulting in a huge number of popover initialization. With this PR only the popover contained in the component is initialized.

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -42,3 +42,6 @@ jobs:
           name: playwright-coverage
           path: coverage-playwright/
           retention-days: 30
+
+    env:
+      ENABLE_COVERAGE: "true"
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -33,3 +33,6 @@ jobs:
           name: unit-coverage
           path: coverage-unit/
           retention-days: 30
+
+    env:
+      ENABLE_COVERAGE: "true"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Unreleased
 
+* Fixed argument popovers performance issue (\#324).
 * Added end to end tests and coverage configuration (\#320).
 
 # 0.6.1
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -21,12 +21,12 @@ To execute the tests seeing the browser add the `--headed` flag or the `--debug`
 Coverage for the unit tests:
 
 ```
-npx vitest run --coverage.enabled --coverage.provider=istanbul --coverage.include=src --coverage.reporter=json --coverage.all
+ENABLE_COVERAGE=true npx vitest run --coverage.enabled --coverage.provider=istanbul --coverage.include=src --coverage.reporter=json --coverage.all
 ```
 
 Warning: coverage takes some time using the option `--coverage.all`.
 
-Coverage for the playwright tests is automatically collected when running `npx playwright test`.
+Coverage for the playwright tests is automatically collected when running `ENABLE_COVERAGE=true npx playwright test`.
 
 To generate a merged html report:
 
--- a/src/lib/components/common/jschema/PropertyDescription.svelte
+++ b/src/lib/components/common/jschema/PropertyDescription.svelte
@@ -3,17 +3,27 @@
 
 	export let description = undefined;
 
-	onMount(() => {
-		const popoverTriggerList = document.querySelectorAll('[data-bs-toggle="popover"]');
-		// eslint-disable-next-line no-undef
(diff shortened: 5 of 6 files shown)


zonia3000:
Some comments about the following message:

```
vite:istanbul> Sourcemaps was automatically enabled for code coverage to be accurate.
 To hide this message set build.sourcemap to true, 'inline' or 'hidden'.
```

Sourcemaps files are usually generated during the build and they are useful for debugging purposes, as well as for the coverage. It looks that istanbul plugin run always, not only during tests. It checks if vite is configured to build the sourcemap files and if it is not it automatically enables that feature.

I added the following to vite.config.js to always enable the building of sourcemaps, since they can be useful also for debugging and I don't think they can have an impact on performance on production code, since they are loaded by the browser only if the developer tools are enabled:

```
build: {
  sourcemap: true
}
```

As a side effect, this also hides that warning.

In any case I don't like the fact that istanbul plugin is always loaded, so I introduced an environment variable `ENABLE_COVERAGE` that has to be set to true to enable istanbul. I updated the pipeline files to have it enabled in our tests.

tcompa:
Thanks @zonia3000!

I'd just need a quick clarification, because we may be using different dictionaries here ;)
Here is mine:
* We load the page corresponding to a given `Workflow` (e.g. the one with ID=8)
* The `task_list` includes a list of `WorkflowTask` objects, each one with its own ID
* Each `WorkflowTask` has a `task` property (which is a `Task` object), and `WorkflowTask.task` also has its own ID
* The `WorkflowTask.task.args_schema` (optional) property is a JSON Schema that describes the N arguments of a given task, possibly in a nested form (an argument may be an array/object itself).
* When we select a given `WorkflowTask` from the `task_list`, a fractal-web component prepares the form to fill/update its arguments, based on `WorkflowTask.task.args_schema`.

You say:
> However it was initializing every time all the popovers of the page for each component, resulting in a huge number of popover initialization. With this PR only the popover contained in the component is initialized.

and this is how I interpret it:
> it was initializing every time all the popovers of the page for all the components (that is, for all the arguments of all `WorkflowTask`s in `task_list`), resulting in a huge number of popover initialization. With this PR only the N popovers related to the currently-selected `WorkflowTask` are initialized.

Am I getting this right?

tcompa:
And thanks for the clarification re:Istabul.

zonia3000:
> > it was initializing every time all the popovers of the page for all the components (that is, for all the arguments of all `WorkflowTask`s in `task_list`), resulting in a huge number of popover initialization. With this PR only the N popovers related to the currently-selected `WorkflowTask` are initialized.
> 
> Am I getting this right?

Yes, I mean these popovers (generated by the `PropertyDescription`  component):

image

Previously each `PropertyDescription` component initialized all the popovers of the page (in the case of the problematic tasks it happened 319 x 319 times).

Now each `PropertyDescription` component initialize only its own component (so in that case we now initialize them 319 x 1 times).

tcompa:
OK, thanks again.

It'd be useful, just for once, to understand where does 319 come from. I'll try to have a simple example, and maybe you can confirm if it makes sense

Here is a simplified JSON Schema, related to the original one:
```json
{
    "title": "CreateOmeZarrMultiplex",
    "type": "object",
    "properties": {
        "allowed_channels": {
            "title": "Allowed Channels",
            "type": "object",
            "additionalProperties": {
                "type": "array",
                "items": {
                    "$ref": "#/definitions/OmeroChannel"
                }
            },
            "description": "DESCRIPTION A"
        }
    },
    "required": [
        "allowed_channels"
    ],
    "definitions": {
        "OmeroChannel": {
            "title": "OmeroChannel",
            "description": "DESCRIPTION B",
            "type": "string"
        }
    }
}
```
and here are some data
```json
{
    "allowed_channels": {
        "key1": [
            "A",
            "B",
            "C"
        ],
        "key2": [
            "D",
            "E",
            "F"
        ]
    }
}
```

Based on the screenshot below, I guess we should be initializing 7 popovers: one for "Allowed Channels" and then six for the items corresponding to A,B,C,D,E,F. And, if I understand correctly, before this PR we would initialize 7x7=49 popovers.
Is this correct?

image

tcompa (approved):

zonia3000:
I didn't retrieved the number in an analytical way. I've just run the following in the browser console:

image

You can expand the first and the last elements and see that they are all popovers from that task arguments.

In any case, your reasoning is correct.

tcompa:
> I've just run the following in the browser console:
> 
> image

Thanks, I confirm it shows 7 nodes in my case (and they can be clearly identified):
image

PR #325, opened 2023-10-25
Task version update

### LINKED ISSUE #285: Update task version from the web interface
EDIT: the most up-to-date description is in a comment below: https://github.com/fractal-analytics-platform/fractal-web/issues/285#issuecomment-1748430756

----

Here's a future feature wish:
Updating to a newer version of a task directly from the web interface for an existing workflow.

Currently, this can be done manually by 
a) Recreating the workflow and setting all parameters again
b) Download a workflow, changing the version of the task in the json, optionally changing parameters that are new, uploading it again.

The vision for this in the web interface would be that it replaces the task with a new version of the task and all the parameters pre-filled. If some parameters of the task changed, give some warning:
- If parameters were added, just list the new parameters
- If some parameters were removed, list the removed parameters and the values that the task had before updating them

This is not urgent, but will be useful when we update tasks versions with patch releases more :)

zonia3000:
## Checklist before merging

Closes #285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
+# Unreleased
+
+* Implemented task version update (\#325).
+
 # 0.6.2
 
 * Fixed peformance issue with argument-description popovers (\#324).
--- a/__tests__/VersionUpdate.test.js
+++ b/__tests__/VersionUpdate.test.js
@@ -79,9 +79,10 @@ describe('VersionUpdate', () => {
 
 		await fireEvent.change(screen.getByRole('combobox'), { target: { value: '2.0.0' } });
 
-		const [moreLink, checkBtn, updateBtnDisabled] = screen.getAllByRole('button');
+		const [moreLink, checkBtn, cancelBtn, updateBtnDisabled] = screen.getAllByRole('button');
 		expect(moreLink.textContent).eq('more');
 		expect(checkBtn.textContent).eq('Check');
+		expect(cancelBtn.textContent).eq('Cancel');
 		expect(updateBtnDisabled.textContent).eq('Update');
 		expect(updateBtnDisabled.disabled).eq(true);
 
@@ -100,9 +101,36 @@ describe('VersionUpdate', () => {
 		});
 		await fireEvent.click(checkBtn);
 
-		const updateBtnEnabled = screen.getAllByRole('button')[1];
+		const updateBtnEnabled = screen.getAllByRole('button')[2];
 		expect(updateBtnEnabled.textContent).eq('Update');
 		expect(updateBtnEnabled.disabled).eq(false);
+	});
+
+	it('use the cancel button when fixing the arguments', async () => {
+		const task = getTask('My Task', '1.2.4');
+		const versions = await checkVersions(task, 1, { changed_property: 'x' });
+		expect(versions[0]).toBe('2.0.0');
+
+		await fireEvent.change(screen.getByRole('combobox'), { target: { value: '2.0.0' } });
+
+		expect(screen.getByRole('textbox').value).eq(
+			JSON.stringify({ changed_property: 'x' }, null, 2)
+		);
+
+		const cancelBtnDisabled = screen.getByRole('button', { name: 'Cancel' });
+		expect(cancelBtnDisabled.disabled).eq(true);
+
+		await fireEvent.input(screen.getByRole('textbox'), {
+			target: { value: '{"changed_property": "y"}' }
+		});
+
+		const cancelBtnEnabled = screen.getByRole('button', { name: 'Cancel' });
+		expect(cancelBtnEnabled.disabled).eq(false);
+		await fireEvent.click(cancelBtnEnabled);
+
+		expect(screen.getByRole('textbox').value).eq(
+			JSON.stringify({ changed_property: 'x' }, null, 2)
+		);
 	});
 
 	it('trying to fix the arguments with invalid JSON', async () => {
--- a/__tests__/VersionUpdate.test.js
+++ b/__tests__/VersionUpdate.test.js
@@ -30,11 +30,15 @@ const task3ArgsSchema = {
 	title: 'MyTask',
 	type: 'object',
 	properties: {
+		changed_property: {
+			type: 'boolean'
+		},
 		new_property: {
 			type: 'string'
 		}
 	},
-	required: ['new_property']
+	required: ['changed_property', 'new_property'],
+	additionalProperties: false
 };
 
 const tasks = [
@@ -70,12 +74,13 @@ describe('VersionUpdate', () => {
 
 	it('update task fixing the arguments', async () => {
 		const task = getTask('My Task', '1.2.4');
-		const versions = await checkVersions(task, 1);
+		const versions = await checkVersions(task, 1, { extra_property: 1, changed_property: 'x' });
 		expect(versions[0]).toBe('2.0.0');
 
 		await fireEvent.change(screen.getByRole('combobox'), { target: { value: '2.0.0' } });
 
-		const [checkBtn, updateBtnDisabled] = screen.getAllByRole('button');
+		const [moreLink, checkBtn, updateBtnDisabled] = screen.getAllByRole('button');
+		expect(moreLink.textContent).eq('more');
 		expect(checkBtn.textContent).eq('Check');
 		expect(updateBtnDisabled.textContent).eq('Update');
 		expect(updateBtnDisabled.disabled).eq(true);
@@ -84,8 +89,14 @@ describe('VersionUpdate', () => {
 			screen.getByText('Following errors must be fixed before performing the update:')
 		).toBeDefined();
 
+		const list = screen.getAllByRole('listitem');
+		expect(list.length).eq(3);
+		expect(list[0].textContent.trim()).eq("must have required property 'new_property'");
+		expect(list[1].textContent.trim()).eq("must NOT have additional property 'extra_property'");
+		expect(list[2].textContent.trim()).contain('/changed_property: must be boolean');
+
 		await fireEvent.input(screen.getByRole('textbox'), {
-			target: { value: '{"new_property": "test"}' }
+			target: { value: '{"changed_property": true, "new_property": "test"}' }
 		});
 		await fireEvent.click(checkBtn);
 
@@ -142,8 +153,8 @@ describe('VersionUpdate', () => {
 	});
 });
 
-async function checkVersions(task, expectedCount) {
-	const result = renderVersionUpdate(task);
+async function checkVersions(task, expectedCount, args = {}) {
+	const result = renderVersionUpdate(task, args);
 
 	// triggers the updates
 	await new Promise(setTimeout);
@@ -159,11 +170,11 @@ async function checkVersions(task, expectedCount) {
 	return options.filter((_, i) => i > 0).map((o) => o.value);
 }
 
-function renderVersionUpdate(task) {
+function renderVersionUpdate(task, args) {
 	const nop = function () {};
 	mockTaskList();
 
-	const workflowTask = { task };
+	const workflowTask = { task, args };
 
 	return render(VersionUpdate, {
 		props: { workflowTask, updateWorkflowCallback: nop, updateNewVersionsCount: nop }
--- a/__tests__/VersionUpdate.test.js
+++ b/__tests__/VersionUpdate.test.js
@@ -79,8 +79,11 @@ describe('VersionUpdate', () => {
 
 		await fireEvent.change(screen.getByRole('combobox'), { target: { value: '2.0.0' } });
 
-		const [moreLink, checkBtn, cancelBtn, updateBtnDisabled] = screen.getAllByRole('button');
-		expect(moreLink.textContent).eq('more');
+		const [moreLink1, moreLink2, moreLink3, checkBtn, cancelBtn, updateBtnDisabled] =
+			screen.getAllByRole('button');
+		expect(moreLink1.textContent).eq('more');
+		expect(moreLink2.textContent).eq('more');
+		expect(moreLink3.textContent).eq('more');
 		expect(checkBtn.textContent).eq('Check');
 		expect(cancelBtn.textContent).eq('Cancel');
 		expect(updateBtnDisabled.textContent).eq('Update');
@@ -92,9 +95,9 @@ describe('VersionUpdate', () => {
 
 		const list = screen.getAllByRole('listitem');
 		expect(list.length).eq(3);
-		expect(list[0].textContent.trim()).eq("must have required property 'new_property'");
-		expect(list[1].textContent.trim()).eq("must NOT have additional property 'extra_property'");
-		expect(list[2].textContent.trim()).contain('/changed_property: must be boolean');
+		expect(list[0].textContent).contain("must have required property 'new_property'");
+		expect(list[1].textContent).contain("must NOT have additional property 'extra_property'");
+		expect(list[2].textContent).contain('/changed_property: must be boolean');
 
 		await fireEvent.input(screen.getByRole('textbox'), {
 			target: { value: '{"changed_property": true, "new_property": "test"}' }
@@ -140,8 +143,9 @@ describe('VersionUpdate', () => {
 
 		await fireEvent.change(screen.getByRole('combobox'), { target: { value: '2.0.0' } });
 
-		const [checkBtn, updateBtnDisabled] = screen.getAllByRole('button');
-		expect(updateBtnDisabled.disabled).eq(true);
+		const checkBtn = screen.getByRole('button', { name: 'Check' });
+		const updateBtn = screen.getByRole('button', { name: 'Update' });
+		expect(updateBtn.disabled).eq(true);
 
 		await fireEvent.input(screen.getByRole('textbox'), {
 			target: { value: '}{' }
--- a/__tests__/VersionUpdate.test.js
+++ b/__tests__/VersionUpdate.test.js
@@ -94,6 +94,25 @@ describe('VersionUpdate', () => {
 		expect(updateBtnEnabled.disabled).eq(false);
 	});
 
+	it('trying to fix the arguments with invalid JSON', async () => {
+		const task = getTask('My Task', '1.2.4');
+		const versions = await checkVersions(task, 1);
+		expect(versions[0]).toBe('2.0.0');
+
+		await fireEvent.change(screen.getByRole('combobox'), { target: { value: '2.0.0' } });
+
+		const [checkBtn, updateBtnDisabled] = screen.getAllByRole('button');
+		expect(updateBtnDisabled.disabled).eq(true);
+
+		await fireEvent.input(screen.getByRole('textbox'), {
+			target: { value: '}{' }
+		});
+		await fireEvent.click(checkBtn);
+
+		expect(screen.getByRole('textbox').classList.contains('is-invalid')).eq(true);
+		expect(screen.getByText('Invalid JSON')).toBeDefined();
+	});
+
 	it('no new versions available for null owner', async () => {
 		const task = getTask('My Other Task', '1.2.3');
 		await checkVersions(task, 0);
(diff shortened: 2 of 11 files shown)


zonia3000:
I added some unit tests at component level:

* update task without changing the arguments
* update task fixing the arguments
* trying to fix the arguments with invalid JSON
* no new versions available for null owner
* no new versions available for admin owner
* display warning if task has no version
* display warning if task has no args_schema

Reported coverage for `VersionUpdate.svelte`: 89.65 % (statements)

tcompa:
Thanks Sonia.

I tested a simple use case (the maximum-intensity-projection task), and everything looks as expected:

Screenshot from 2023-10-25 14-38-53
Screenshot from 2023-10-25 14-38-59
Screenshot from 2023-10-25 14-39-04
Screenshot from 2023-10-25 14-39-09
Screenshot from 2023-10-25 14-39-12

And great for the extensive coverage!

I want to also try out with a more complex test

tcompa:
Minor glitch: the success message should be something like "Task version updated successfully" (rather than "Workflow updated")

tcompa:
I ran some more extensive tests, directly from the fractal-web interface.

I created 5 version of the same task, with the following JSON schemas:
```json
// SCHEMA 1
{
    "title": "Schema1",
    "type": "object",
    "properties": {
        "property1": {
            "title": "Property 1 (array)",
            "type": "array",
            "items": {
                "type": "string"
            }
        }
    },
    "required": [
        "property1"
    ],
    "additionalProperties": false
}

// SCHEMA 2: add required property2
{
    "title": "Schema2",
    "type": "object",
    "properties": {
        "property1": {
            "title": "Property 1 (array)",
            "type": "array",
            "items": {
                    "type": "string"
            }
        },
        "property2":{
            "title": "Property 2 (string)",
            "type": "string"
        }
    },
    "required": [
        "property1", "property2"
    ],
    "additionalProperties": false
}

// SCHEMA 3: remove property1
{
    "title": "Schema3",
    "type": "object",
    "properties": {
        "property2":{
            "title": "Property 2 (string)",
            "type": "string"
        }
    },
    "required": [
        "property2"
    ],
    "additionalProperties": false
}

// SCHEMA 4: change type of property2
{
    "title": "Schema4",
    "type": "object",
    "properties": {
        "property2":{
            "title": "Property 2 (boolean)",
            "type": "boolean"
        }
    },
    "required": [
        "property2"
    ],
    "additionalProperties": false
}


// SCHEMA 5: add optional properties with/without default values
{
    "title": "Schema5",
    "type": "object",
    "properties": {
        "property2":{
            "title": "Property 2 (boolean)",
            "type": "boolean"
        },
        "property3":{
            "title": "Property 3 (string, no default)",
            "type": "string"
        },
        "property4":{
            "title": "Property 43 (string, with default)",
            "type": "string",
            "default": "default4"
        }
    },
    "required": [
        "property2"
    ],
    "additionalProperties": false
}
```
and then I sequentially updated a task in a workflow, from version 1 to 5 - see some screenshots below.

Overall, I could not spot unexpected behaviors. The only change I would definitely recommend is to make the JSON-text field **much** larger.

To be honest, I think that when a user has to modify the JSON directly a bunch of things can go wrong, but in principle this is some kind of advanced use case - while the typical `fractal-tasks-core` task-version updates should rarely include new required arguments.

---

Screenshot from 2023-10-26 09-16-46
Screenshot from 2023-10-26 09-17-04
Screenshot from 2023-10-26 09-17-19
Screenshot from 2023-10-26 09-17-28

tcompa:
@jluethi a second review woud be useful here. For instance: do you think we need to include more instructions for users and/or descriptions of what will happen or just happened?

If you want to start with my example, here is the sqlite DB I prepared with the five tasks mentioned above: [test.db.txt](https://github.com/fractal-analytics-platform/fractal-web/files/13174219/test.db.txt).

tcompa:
Something else I checked: if there exists a new version of the task, but it has no `args_schema`, then this is not included in the update candidates.

This is as expected.

jluethi:
That's great to hear!

> To be honest, I think that when a user has to modify the JSON directly a bunch of things can go wrong, but in principle this is some kind of advanced use case - while the typical fractal-tasks-core task-version updates should rarely include new required arguments.

Agreed, but let's still make it as simple as we can, because updating tasks is an action that will occur often. And we don't want to constrain ourselves to basically never change parameters because it would get too hard.

In that spirit: What is shown to be edited by the user in the "Fix the arguments" box? Is it only the argument that needs changing? Or the whole schema? Agreed with a bigger box :)

And is there a way to sanitize the error message a bit? If the message also included which property is the issue, we could just show the message instead of the whole json, no?

jluethi:
> @jluethi a second review woud be useful here. For instance: do you think we need to include more instructions for users and/or descriptions of what will happen or just happened?
> 
> If you want to start with my example, here is the sqlite DB I prepared with the five tasks mentioned above: [test.db.txt](https://github.com/fractal-analytics-platform/fractal-web/files/13174219/test.db.txt).

Thanks! Will try to test by tomorrow to give more informed feedback! :)

tcompa:
> In that spirit: What is shown to be edited by the user in the "Fix the arguments" box? Is it only the argument that needs changing? Or the whole schema?

For the moment this is the whole `WorkflowTask.args` objects. That is, we are not re-creating a custom form that knows about the old/new JSON Schemas, but just asking to update the JSON string directly. After the user fills it, the "check" button performs the validation, and then the "update" button gets enabled.

My take on this: editing task arguments is where the user spends a lot of time, and it's worth a custom (or maybe not custom, in the future) component that converts the requirements (i.e. the JSON Schema) into a user-friendly form. I'm not sure that the "update from one schema to another one" feature is worth the same complexity; for the moment I would say it doesn't.

> And is there a way to sanitize the error message a bit? If the message also included which property is the issue, we could just show the message instead of the whole json, no?

I'm not sure it's easy to do it in a way which still covers any kind of ajv validation error, but I'll let @zonia3000 chime in.
Possibly relevant links:
* https://www.npmjs.com/package/ajv-errors
* https://www.npmjs.com/package/@apideck/better-ajv-errors

tcompa:
> For the moment this is the whole `WorkflowTask.args` objects. That is, we are not re-creating a custom form that knows about the old/new JSON Schemas, but just asking to update the JSON string directly. After the user fills it, the "check" button performs the validation, and then the "update" button gets enabled.
> 
> My take on this: editing task arguments is where the user spends a lot of time, and it's worth a custom (or maybe not custom, in the future) component that converts the requirements (i.e. the JSON Schema) into a user-friendly form. I'm not sure that the "update from one schema to another one" feature is worth the same complexity; for the moment I would say it doesn't.

If we want to follow the more complex&user-friendly direction, let's first explore existing tools (e.g. https://www.npmjs.com/package/json-schema-diff), before writing a new custom component like we did for the "editing task arguments" feature.

zonia3000:
In general, it's not so easy to reduce ajv errors to a single human friendly string, since there are many cases to be considered and the library can generate a lot of different parameters in the messages.

I had a look to these packages:

> * https://www.npmjs.com/package/ajv-errors
> * https://www.npmjs.com/package/@apideck/better-ajv-errors

`ajv-errors` requires to modify the schema adding some custom error messages on it, so it's not what we need in this case.

`better-ajv-errors` doesn't return a complete message. Looking at the second example (enum validation) the message is saying that "property must be equal to one of the allowed values", but the list of allowed values has to be retrieved from a different field. So the messages should still be rearranged even after the rewriting done by this package, otherwise we will loose some information.

Moreover, the library itself doesn't contain a lot of logic. Most of the code is contained in this 120 lines file: https://github.com/apideck-libraries/better-ajv-errors/blob/main/src/index.ts

I would prefer to write my custom function considering the following 3 cases:

1. a required property is missing
2. an extra property is found
3. a property is not valid (wrong type, value too big, value not allowed, ...): in this case I would add a link that can be clicked to expand the item showing the complete message, so that a user can decide to know more about the error only if needed.

An alternative, as suggested by @tcompa, would be to generate a form for this situation evaluating the difference between the two JSON schemas. Let me know what do you prefer and also if you would like that I start working on this now or if it is better to plan it as an enhancement to be done in a future task.

tcompa:
Side comment: a tiny step in the user-friendly direction would be to add JSON syntax highlighting to the text input prompt.
(this is unrelated to the ongoing discussion on whether a custom form is needed)

zonia3000:
An example of what I've described before:

image

Expanded version:

image

jluethi:
I like the custom handling you're showing here, with a fallback to showing the full error if it's not one of the simpler (and likely most common) cases!

tcompa:
I am now testing the latest version, and I have a few minor remarks. I'll write them in single comments, to simplify the discussion.

tcompa:
1. I propose we add a "cancel" button to the update procedure, which brings us back to the original state at any point before a successful "update"

tcompa:
2. There is a use case which we are not covering, that is the case where task arguments are not even valid under the "old" JSON schema (this happens every time I add a task which has required arguments which are not set by default). If I try to update such a task, I will get an error which combines the requirements of the old and new schemas, like:
image
I think it would be useful to add a constraint: if the current task doesn't have valid arguments under the old schema, you cannot start the update-version procedure.

tcompa:
As we discussed above, providing user-friendly error messages is tricky. For instance the message below is not very clear, since it does not tell the user what is the type of the missing property. 
 
image

One way to find it out is to guess it and then "check" again - so that the next error is more explicit:

image

Proposed change: can we *always* include the "more" button that expands the full error message? Maybe this can help in finding useful information.

tcompa:
If it's not too complex, could we add JSON syntax highlighting to the text in the "fix the arguments" input block?

tcompa:
Last remark is not about the implementation, but more about the use cases where this could lead to unexpected behavior.

What happens when a fractal-tasks-core version (e.g. v0.12.0) introduces breaking changes to the internal logic of some tasks (with no changes to the arguments' schema)? In the current situation, the user would always be "encouraged" to always update, since they see a "version **(1)**" tab title. This would immediately lead to a broken workflow, which cannot run any more until further actions are taken (e.g. repeating the image-to-zarr conversion).

Should we somehow make it harder to have this unexpectedly-breaking updates?
Some projects have very structured docs, which say e.g. when a certain API element was modified and what changed, but we won't be able to do the same in an automated way.
I see a couple of possible ways to mitigate the problem:
1. Whenever the new version changes the minor (or major) release (e.g. from 0.11.2 to 0.12.0), we show a warning on the line of "This change could be breaking, please review the CHANGELOG of the task package or ask the mantainers". If task-package maintainers are somewhat careful in their use of semantic versioning, and if there is easy access to the CHANGELOG, then this could be useful.
2. We start to list breaking changes in the docstrings of tasks, and in fractal-web we display a comparison of the old and new task docstrings before the update-version procedure.

I tend to think that option 1 is the more interesting one, at least for fractal-tasks-core tasks. Especially if we are able to add a direct link to the CHANGELOG docs page. Option 2 seems a bit too heavy on the fractal-tasks-core side (we would need to enforce this new rule on docstring, and they may get very verbose at some point), and also it would display a lot of (mostly useless) information to the user.

Of course option 3 also exists: we always display a generic warning, and let the user do whatever they want.

Feedback or other ideas are welcome (cc @jluethi)

zonia3000:
> Proposed change: can we _always_ include the "more" button that expands the full error message? Maybe this can help in finding useful information.

It's ok for me to include always the "more" link, probably it is better to be more homogeneous, but consider that in case of the required message the complete JSON doesn't contain information about the type of the property:

image

jluethi:
I'll need to test myself as well to be able to give better answers and will try so soon.

In the meantime:

> If it's not too complex, could we add JSON syntax highlighting to the text in the "fix the arguments" input block?

Sure, nice if possible.

> What happens when a fractal-tasks-core version (e.g. v0.12.0) introduces breaking changes to the internal logic of some tasks (with no changes to the arguments' schema)?

My first intuition: Out of scope for the core feature. The main problem we want to solve is updating tasks when they don't change backwards compatibility. We need to avoid these breaking changes whenever possible on the task side. And when we do them, it needs to come with communication to users.
I don't mind option 1, but I doubt users would be able to do too much with it. It's nice to link to the changelog so that users can see whether the update brings a fix they were looking for. Anything beyond that anyway lands back with users asking support (i.e. me) ;)

How much should we worry about it?
If we make it easy to export a workflow based on job history, users could always go back to an old workflow with all the old task versions later. But we want to encourage users to find good ways to make things run on the more recent versions of the tasks

tcompa:
> It's ok for me to include always the "more" link, probably it is better to be more homogeneous, but consider that in case of the required message the complete JSON doesn't contain information about the type of the property:

Ah, too bad ;)

First, I agree: let's always include the "more" link, for the sake of homogeneity.

---


Second, if we want to add more context we would need some customization.
In some cases that would be "easy": we read the `schemaPath` property of the ajv error, we load that specific path from the new JSON Schema, and we show the user what are the JSON-Schema specific requirements. But this is not general, because it already fails for https://github.com/fractal-analytics-platform/fractal-web/pull/325#issuecomment-1782448416 (`"schemaPath": "#/required" would not immediately load the `prop2` JSON Schema component).

Then we are back at Sonia's comment above:
> I would prefer to write my custom function considering the following 3 cases:
> 1. a required property is missing
> 2. an extra property is found
> 3. a property is not valid (wrong type, value too big, value not allowed, ...): in this case I would add a link that can be clicked to expand the item showing the complete message, so that a user can decide to know more about the error only if needed.

That is, we would need to define a set of cases (e.g. the three above) which we handle in a custom way. For instance we can follow the JSON-Schema references and find out what is the type of a missing property, and tell the user "Missing `prop2` property, of type `string`". For anything that is not part of "known" cases, we would just print the ajv message, and have the "more" link to expand the raw error.

tcompa:
> > What happens when a fractal-tasks-core version (e.g. v0.12.0) introduces breaking changes to the internal logic of some tasks (with no changes to the arguments' schema)?
> 
> My first intuition: Out of scope for the core feature

> I don't mind option 1, but I doubt users would be able to do too much with it. It's nice to link to the changelog so that users can see whether the update brings a fix they were looking for. Anything beyond that anyway lands back with users asking support (i.e. me) ;)

Even option 1 does not come for free, as we first need to modify the manifest and manifest schema in fractal-tasks-core and in fractal-server. If your intuition is that this is out-of-scope for now, then I'd say let's keep it as a possible future improvement.

jluethi:
> Even option 1 does not come for free, as we first need to modify the manifest and manifest schema in fractal-tasks-core and in fractal-server. If your intuition is that this is out-of-scope for now, then I'd say let's keep it as a possible future improvement.

Agreed, let's keep this as a future improvement in this case.

jluethi:
I now tested the behavior. Overall:
Great, very nice that we can update tasks this way!

In the (hopefully rare) cases where the user gets send to the schema editing:
I like the "check" functionality and the additional feedback it provides. It's not very obvious what to do initially though. e.g. here
Screenshot 2023-10-27 at 10 25 27

The user wouldn't know what type is expected. They only get that feedback when they "check". But again, I'd say fine for a first implementation, it does work!

-----
In general, I think we have a working v1 and can improve some updating behaviors in the future. 

For example, I'd ask the higher-level question: When do we actually need to expose the full json args schema? Before we find to many fancy ways of handling this, I would suggest exploring a solution in the following direction:
Can we handle the argument missing or argument removed cases even more specific? For example:
- A new argument is required: Fine, we update the task and a parameter is missing. The user gets sent to the args editing interface and can only save once they put in the parameter.
- A parameter needs to be removed? We'll show a warning that this parameter will be removed. User clicks ok, then it continues


And for everything more complex, we fall back to the current behavior :)

jluethi:
> That is, we would need to define a set of cases (e.g. the three above) which we handle in a custom way. For instance we can follow the JSON-Schema references and find out what is the type of a missing property, and tell the user "Missing prop2 property, of type string". For anything that is not part of "known" cases, we would just print the ajv message, and have the "more" link to expand the raw error.

I also like this idea for future improvements. But let's keep this in mind for v2 of updating, together with my other ideas above. For the v1 of adding task updates, this isn't required yet

jluethi:
tl;dr:
Let's not dive too much deeper into the complex cases of error handling for complex task updates with changing required parameters for now.

I think adding a link to where to find what has changed in versions (maybe link to `Docs Link` that we also have in info?) on the page would still be helpful.

I will test it with the actual fractal-tasks-core tasks next to get some idea there :)

tcompa:
Here's a recap, if I got everything right..

Missing features:
1. Cancel button (https://github.com/fractal-analytics-platform/fractal-web/pull/325#issuecomment-1782385169)
2. JSON syntax, if not too complex (https://github.com/fractal-analytics-platform/fractal-web/pull/325#issuecomment-1782388495)

To decide:

3. Should we add a generic warning that is always there, on the line of "Information on different version may be found at {task.docs_link}"?
4. Do we care about the case where both requirements of the old and new schemas need fixing (see https://github.com/fractal-analytics-platform/fractal-web/pull/325#issuecomment-1782386170)? It's an edge case (it means that nobody ever ran that workflow), and then I'm fully OK with ignoring it.

Future:

5. Anything related to providing either better errors or smoother ways to get to fix the arguments.

jluethi:
I have now tested the fractal-tasks-core package, from 0.10.0 all the way through 0.13.1 (incrementally and all at once).

Main takeaways: This version is ready! I was only sent to the json editor twice for minor changes, because most parameter changes aren't for required parameters.
The functionality is great and for the relatively rare cases of having to interact with the json schema, the current complexity there is fine! :)

The only thing I'd add to v1 of this: Some way of highlighting that a workflow has tasks that aren't up-to-date even if one isn't on that workflow task page. I.e. if a workflow looks like this:
Screenshot 2023-10-27 at 11 06 57

The current task doesn't have updates, but some others do. Should we show some little yellow exclamation mark at the top next to the workflow name, that shows "some tasks are not on the most recent version" on hover? Or a little yellow exclamation mark/a number next to each task that isn't up to date in the task list on the left?

jluethi:
> Cancel button

Is this required? isn't leaving the page just cancelling the update? e.g. if you don't click "Update" at the bottom, it doesn't actually update?

> Should we add a generic warning that is always there, on the line of "Information on different version may be found at {task.docs_link}"?

I think that's the correct level. With the benefit of also giving context of what the version would be. Any more complex warning => let's see whether we need to get into that.

> Do we care about the case where both requirements of the old and new schemas need fixing (see https://github.com/fractal-analytics-platform/fractal-web/pull/325#issuecomment-1782386170)? It's an edge case (it means that nobody ever ran that workflow), and then I'm fully OK with ignoring it.

I would say no. You are supposed to have arguments that pass under the current version, otherwise you get that error. Further improvements on this would be for a future version


Additional point from me above:
6. Some indicator for the workflow that some tasks have updates

tcompa:
> > Cancel button
> 
> Is this required? isn't leaving the page just cancelling the update? e.g. if you don't click "Update" at the bottom, it doesn't actually update?

If you refresh the page (or switch to another one), then the update is cancelled
If you select a different task, then the update is cancelled
If you select the "Arguments" tab and then go back to "Version", then the update is still there

zonia3000:
I would suggest to postpone the JSON syntax highlighting. Maybe at the next iteration we could directly implement the JSON diff approach and have a user friendly form without having to install extra dependencies.

I tried this: https://github.com/touchifyapp/svelte-codemirror-editor

It works fine, however it breaks all the unit tests. It seems that it is not possible to retrieve the textarea inside the unit tests using this component. I only see the wrapper div, with an empty content.

I tried also with this: https://github.com/nateshmbhat/svelte-ace
But it leads to some vite errors during the build.

tcompa:
I updated the recap here: https://github.com/fractal-analytics-platform/fractal-web/pull/325#issuecomment-1782565259

tcompa:
> I would suggest to postpone the JSON syntax highlighting. Maybe at the next iteration we could directly implement the JSON diff approach and have a user friendly form without having to install extra dependencies.

OK. Thanks for checking!

zonia3000:
I've changed a bit the text of the general warning message, using "task documentation" instead of the complete URL because it may be quite long:

image

Tasks with available new versions have a warning sign directly in the list:

image

I've also added the cancel button and added the "more" link to all the errors.

tcompa:
I checked once again, and could not spot anything unexpected.

Very last suggestion: could the yellow warning be placed above the red error? IMO the warning should be close to the version-choice dropdown menu, while the error message should be close to the JSON editor.

image

Other than that, I've nothing more to add.

tcompa:
As of https://github.com/fractal-analytics-platform/fractal-web/pull/325/commits/688168412ea8833e058de62eb738d4cbdc0a0d0d (thanks!), I'd say we are good for merge&release

jluethi:
This is awesome!! 🎉

PR #328, opened 2023-10-31
Created per-user and per-workflow jobs pages

### LINKED ISSUE #119: Automatic updating of the job page to fetch new status
e.g. every x seconds

Fancy: Only refresh while any workflow is on running

### LINKED ISSUE #294: Always sort by jobs starting from latest start-time
Does the job page have a default sorting? If so, I can't figure out what it is.

It would make sense to sort by default by ID/Start Time (the sorting for both should be equivalent). Newest job should be on top.

### LINKED ISSUE #318: New per-user and per-workflow job pages
Let's move to two different "jobs" pages: one at the workflow level, and one with all jobs of a user

## Per-workflow jobs
1. This fetches all jobs related to a workflow, by fetching all jobs related to the corresponding project and then filtering on fractal-web side. Maybe we can improve this based on https://github.com/fractal-analytics-platform/fractal-server/issues/920
2. Data are filtered _before_ filling the table, so that the table at first appears as clean (with all filters unset) and the URL does not include any query parameter
3. The header of the page correctly says something like `Jobs of workflow "My Workflow Name"`
4. The url&breadcrumbs of the page are appropriate (something like `project/1/workflow/2/jobs` in the URL, and the same structure in the breadcrumbs)
5. A (top-right?) button brings you back to the workflow page
6. The table does not include project and workflow columns

## Per-user jobs

1. It is always visible in the top menu, under the "Jobs" name
2. It fetches all user's jobs, by first fetching all user's projects and then fetching all jobs of each project. Maybe we can have a single endpoint for this? (https://github.com/fractal-analytics-platform/fractal-server/issues/920)
3. It does not apply any filter on the job list
4. It does show project.name and workflow.name columns
5. The header of the page correctly says something like `Jobs of user "admin@something.xxx"`

## Side effects

Other simplifications that are now possible
1. The workflow-apply modal, upon success, redirects to the workflow page (without adding the job-id query parameter to the URL)
2. We can probably fully disable the feature of URL query parameters being turned into table filters
3. We automatically get rid of the "Job ID filters don’t show up in the interface" issue
4. We most likely get rid of the "Transient display of other jobs during refresh: " issue

## Related

* #119 
* #294

### LINKED ISSUE #331: Bring back playwright to stable versions
We will provisionally use `next`, and this is just a reminder.

### LINKED ISSUE #337: Add page-loading spinner
(as discussed with @zonia3000)

We do need to better profile the current loading times, and discuss mitigation strategies (pagination, slim response models, ..), but in terms of usability it's nevertheless important to add a spinner for when loading a page takes "long".

zonia3000:
## Checklist before merging

The two pages are based on the same component (`JobsList`), with different setup.

Closes #318
Closes #294
Closes #119
Closes #337
Closes #331


zonia3000:
I opened as draft since I would like to implement also #119 here. The 2 different job pages can be tested in the meanwhile.

tcompa:
Here is a slightly unexpected behavior.

I have a user with no project.

I head to http://localhost:5173/jobs through the "Jobs" link, and it shows me an empty page (not a page with an empty table):
image.

Then I refresh the page, and I get a 500 error. I guess it is this error from the svelte log, rather than a 500 from fastapi. But I'll need to double-check.
```
fractal-web     | TypeError: Reduce of empty array with no initial value
fractal-web     |     at Array.reduce (<anonymous>)
fractal-web     |     at file:///app/fractal-web/build/server/chunks/JobsList-4c2a174d.js:826:52
fractal-web     |     at Object.$$render (file:///app/fractal-web/build/server/chunks/index3-eb2fa167.js:123:18)
fractal-web     |     at file:///app/fractal-web/build/server/chunks/_page.svelte-a92c3513.js:32:45
fractal-web     |     at Object.$$render (file:///app/fractal-web/build/server/chunks/index3-eb2fa167.js:123:18)
fractal-web     |     at Object.default (file:///app/fractal-web/build/server/index.js:89:97)
fractal-web     |     at file:///app/fractal-web/build/server/chunks/_layout.svelte-5765ba14.js:33:60
fractal-web     |     at Object.$$render (file:///app/fractal-web/build/server/chunks/index3-eb2fa167.js:123:18)
fractal-web     |     at file:///app/fractal-web/build/server/index.js:54:101
fractal-web     |     at $$render (file:///app/fractal-web/build/server/chunks/index3-eb2fa167.js:123:18)
```

zonia3000:
> Here is a slightly unexpected behavior.
> 
> I have a user with no project.
> 
> I head to http://localhost:5173/jobs through the "Jobs" link, and it shows me an empty page (not a page with an empty table)
> 

Thanks, I've fixed this.

zonia3000:
I did some improvements but it's still quite slow when using the test database because it has many jobs having null `workflow_dump`. I would wait [#986](https://github.com/fractal-analytics-platform/fractal-server/issues/986) before considering this completed.

tcompa:
> I did some improvements but it's still quite slow when using the test database

Agreed. I was just testing the latest commit, and indeed it's quite slow.

> I would wait [#986](https://github.com/fractal-analytics-platform/fractal-server/issues/986) before considering this completed.

Now released in fractal-server 1.4.0a3.

tcompa:
Side comment: I do appreciate the improvement in column formatting introduced with https://github.com/fractal-analytics-platform/fractal-web/pull/328/commits/dc1a19cb3ec53576b50752b9b6924256b65add15.

tcompa:
Based on our discussion related to #338, the current view of the user's jobs page is like
image

The "undefined" values are expected, and will be automatically fixed by updating the DB (ref https://github.com/fractal-analytics-platform/fractal-server/issues/1010).

tcompa:
Re #294 

In my test, I confirm that jobs are sorted as expected:

image

tcompa:
A minor issue related to #119:
When I am at http://localhost:5173/projects/60/workflows/70/jobs, the background refresh calls go to http://localhost:8000/api/v1/project/60/job/, but I would expect them to use http://localhost:8000/api/v1/project/60/workflow/70/job/.

tcompa:
A quick check, based on #318:

> We can probably fully disable the feature of URL query parameters being turned into table filters

Is this currently done? That is, have we fully removed the feature of URL query parameters (like http://localhost:5173/projects/60/workflows/70/jobs?job_id=123)?

tcompa (approved):

zonia3000:
> A quick check, based on #318:
> 
> > We can probably fully disable the feature of URL query parameters being turned into table filters
> 
> Is this currently done? That is, have we fully removed the feature of URL query parameters (like http://localhost:5173/projects/60/workflows/70/jobs?job_id=123)?

I'm not handling the query parameters anymore and I've removed them from the redirect that happens when the job starts. Do you see them somewhere?

tcompa:
> I'm not handling the query parameters anymore and I've removed them from the redirect that happens when the job starts. Do you see them somewhere?

All good then. No, I didn't see them anywhere. I only wanted to double check.

PR #350, opened 2023-11-29
Used new endpoints for retrieving current user and list of users

### LINKED ISSUE #335: Align with fractal-server 1.4.0
**[this issue is still in progress]**

Main changes in fractal-server 1.4:

1. There is an important change which does not require immediate action in fractal-web but that we'll need to keep in mind. It concerns the ApplyWorkflow (AKA "Job") table, where the four existing foreign keys (input_dataset_id, output_dataset_id, workflow_id, project_id) become optional. This means that from now on there will be different types of jobs:
    * A job with all FKs non-null, which is in relation with a workflow, a project, and one or two datasets
    * A job which is still in relation with a project (and therefore accessible to the user), but possibly not with a workflow (or with a dataset). This is the typical situation when a workflow (or a dataset) is removed after the job already exists.
    * A job which has all four FKs set to null, which happens when the user deletes a whole project. This kind of jobs are meant to be an archive, which is only accessible to a superuser for accounting reasons.
2. In some cases, when looking at jobs which have some null FK it will be useful to learn something about the workflow or dataset(s) that they were related to at creation time. This will be possible via the following job properties: `workflow_dump`, `input_dataset_dump`, `output_dataset_dump`, that are (obviously) dumps of the corresponding DB items taken at the job-creation time.
3. In `GET /api/v1/project/` and `GET /api/v1/project/{project_id}`, project acquires new properties `workflow_list` and `job_list` (while `dataset_list` was already present in previous versions).
4. There will be a new set of superuser-restricted monitoring endpoints (https://github.com/fractal-analytics-platform/fractal-server/pull/947), to be used as part of https://github.com/fractal-analytics-platform/fractal-web/issues/301.
5. There will be a couple of new endpoints for getting jobs at the per-workflow and per-user level (https://github.com/fractal-analytics-platform/fractal-server/issues/920)

I'll continue later, and include more specific details of things that may require changes in fractal-web.

zonia3000:
## Checklist before merging

Closes #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 This release requires fractal-server 1.4.0.
 
+* Used new endpoints for retrieving current user and list of users (\#350).
 * Added user profile page (\#336).
 * Added admin area with users management (\#336).
 * Added Jobs button in home page (\#346).
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -38,7 +38,7 @@ export default defineConfig({
 
 	webServer: [
 		{
-			command: './tests/start-test-server.sh 1.4.0a6',
+			command: './tests/start-test-server.sh 1.4.0a8',
 			port: 8000,
 			waitForPort: true,
 			stdout: 'pipe',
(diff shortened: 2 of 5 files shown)


tcompa:
Thanks Sonia!
I used this a bit with fractal-server 1.4.0a8, and could not spot any unexpected behavior. I'm merging this one, and let's open issues if needed.

Issue #347, opened 2023-11-28
Let admin change status of a job


tcompa:
The corresponding PATCH endpoint will be available at `/api/v1/job/{job_id}/` as of
* https://github.com/fractal-analytics-platform/fractal-server/pull/1030

I guess the best place for this option is within a "Options" column in the upcoming admin/jobs page (ref https://github.com/fractal-analytics-platform/fractal-web/issues/301)

tcompa:
The `PATCH /api/v1/job/{job_id}/` endpoint is available e.g. in fractal-server 1.4.0a8, and for the moment the only accepted payload is
```json
{
  "status": "failed"
}
```

tcompa:
This was added in #352.

PR #352, opened 2023-11-30
Admin jobs page

### LINKED ISSUE #301: Admin tools
Based e.g. on #9, and for instance:

* Display user's project list (user.project_list), with links to the corresponding project pages
* Modify a user's project list (e.g. to associate a user to an existing project) -> not yet exposed in fractal-server
* Display user-related jobs
* All fractal jobs
* ...

### LINKED ISSUE #353: Handle empty job list for a workflow
Starting from a project page
image
I click on "List jobs", in the workflow list.
This leads to an empty page:
image

I think it'd be better to show one of:
1. An empty table
2. A message saying something like "There is no job associated to the current project".

zonia3000:
## Checklist before merging

Closes #301
Closes #353


tcompa:
> A note about the date time pickers inside the jobs form:

Two comments:
1. After a first test, I think there's nothing wrong with the one you picked.
2. If there's a place where we want the UI to shine, I think it is not the admin area but rather what the common users see ;)

TL;DR the option you chose is perfectly fine IMO

tcompa:
I've gone through the new admin-area job page a bit, and it looks nice!

I think we could use "Admin area" everywhere (also in the top-menu), rather than "Administration".

Here are a couple of comments about the "options" column:

1. The `/api/v1/project/1/job/1/download/` and `/api/v1/project/1/job/1/stop/` endpoints will return a 403 for jobs that do not belong to the admin. We can easily modify them in fractal-server to always make it possible for a superuser to call them, but they are currently not available. For the moment I suggest we remove the two corresponding buttons, and then take decisions on what an admin can do: there's nothing wrong in downloading logs, but should an admin be able to cancel a user's job?
2. I suggest we also include https://github.com/fractal-analytics-platform/fractal-web/issues/347. Note that that endpoint will *not* cancel a job execution, but it's only meant to clean up a wrong `job.status` in the database. For this reason, it should not be exposed through a "cancel" button, but rather through an "edit" button. Then the admin can set a job status to `failed` from within a (admittedly very simple) "JobEdit" modal.

(context for these remarks: the main goal of this page is for the admin to _view_ what's going on. We can definitely also expose some active operations, but that's not the first goal)

tcompa:
Here are a few of screenshot of the current status, which I think do match with expected behavior.

Note that we are exposing the choice of user as a pre-filled drop down menu, which works because we wont' have more than a few tens of them for the moment. The same cannot be done for IDs of projects/workflows/datasets, since there would be too many. For the moment, if the admin really needs job of a given workflow then they need to use the workflow ID. We can improve on this in the future, but I think it's not the highest priority.

image

image

image

image

image

tcompa:
One more minor comment, related to https://github.com/fractal-analytics-platform/fractal-web/pull/352/commits/5b02c75a5b769613609b43acc3993fbcf8d68f7e (that is, the download-csv feature).

It would be useful to add some more columns:
* `working_dir`
* `working_dir_user`
* `first_task_index`
* `last_task_index`

first_task_index

tcompa:
The same access issue as in https://github.com/fractal-analytics-platform/fractal-web/pull/352#issuecomment-1835634553 also applies to the "refresh" button within the "info" modal:

image

jluethi:
Not a full review, just some short comments on the cool things I'm seeing and reading here:

Overall, looks super promising & valuable!

> For the moment I suggest we remove the two corresponding buttons, and then take decisions on what an admin can do: there's nothing wrong in downloading logs, but should an admin be able to cancel a user's job?

Would be nice to have those 2 eventually :)

> Note that we are exposing the choice of user as a pre-filled drop down menu, which works because we wont' have more than a few tens of them for the moment. The same cannot be done for IDs of projects/workflows/datasets, since there would be too many. For the moment, if the admin really needs job of a given workflow then they need to use the workflow ID. We can improve on this in the future, but I think it's not the highest priority.

That seems totally fine! I think it would mostly be a filter by user to start with. A next step may be to easily expose seeing the project list with their job IDs. We can have an admin use IDs :)


> the project names remain undefined

Seeing the project undefined: Could we show project ID there in the table? (even if it wasn't clickable)

tcompa:
> > there's nothing wrong in downloading logs, but should an admin be able to cancel a user's job?
> 
> Would be nice to have those 2 eventually :)

OK. This requires changes on fractal-server - I'll open an issue over there.

---

Just to clarify: what I am not comfortable with is to have the same "cancel" button (in the "options" column) as in the user's jobs page. The admin jobs page will eventually include plenty of jobs, and IMO it's too risky to be one click away from stopping one of them. Note that the corresponding risk in the user's pages is still present, but less severe because:

1. There are less rows in the table, and then clicking the wrong one is less likely
2. A user can only delete their own jobs

tcompa:
As for "clickability", the current situation is that all links to projects/workflows/datasets owned by a user who is not the current superuser will be broken. I think we should fix it (in a future PR) by providing a set of consistent browsing options which only point to the admin-area. For instance instead of linking the project name to http://localhost:5173/projects/999 we could point to http://localhost:5173/admin/projects/id=999 - or something similar.

The idea is that as long as we remain in the admin area and in the `/admin` API block , all links are valid and all API calls are authorized. If we exit from this area, we end up in the `/projects` pages and in the `/api/v1` endpoints - which are not currently usable.

tcompa:
A minor issue:
I'd rather not show the pencil button next to jobs which are in "done" status, but only for "submitted" or "running" ones.

tcompa:
Another remark:

For jobs that do not belong to the admin, we do have access to the project name - as is visible in the info modal:

image

However, we see a `-` field in the main table:
image

Issue #90, opened 2023-04-05
Show dynamic job status in workflow page


tcompa:
Some complexities:
* Multiple jobs for the same workflow but different input datasets.
* Granular per-task status
* Resubmission or running from task A to task B (with which I/O dataset?)

tcompa:
A new endpoint was introduced in fractal-server 1.3.5 (and refined later, e.g. in 1.3.12): `GET /api/v1/project/{project_id}/dataset/{dataset_id}/status`. The response looks like:
```json
{
  "status": {
    1: "submitted",
    2: "failed",
    3: "done"
  }
}
```
where 1, 2 and 3 are WorkflowTask IDs.


Here is a first proposal of how we can expose this endpoint _for testing_ (*):
1. For each job (in the jobs-page table), expose a button "Check status"
2. Once clicked, call the new `status` endpoint, with `dataset_id=job.output_dataset_id`
3. The response includes all WorkflowTask's which have been associated to that dataset, but those can also be part of other jobs or other workflows. We then need to extract the subset of WorkflowTasks that belong to the current job. This is done by looking at the IDs of the `job.workflow_dump.task_list` (to be precise, we should only look at the sub-array defined by the positional indices `job.first_task_index` and `job.last_task_index`.
4. We now have a list of the WorkflowTask's that are part of the current job, and of their statuses. As a first implementation, we can show this information as a simple three-columns table: WorkflowTask.id, WorkflowTask.task.name, status.


(*) The longer-term perspective is to integrate this information directly in the Workflow page, but here I'm only proposing a first way to play with this new feature and verify that its response matches with expectations. Note that the longer-term perspective above also requires establishing a link between a Workflow and a Dataset.

jluethi:
Great, this sounds like a good way to start testing the status feature!

jluethi:
I've sketched some scenarios for monitoring and how this can be displayed on the workflow page.
<img width="1014" alt="Screenshot 2023-11-24 at 11 43 54" src="https://github.com/fractal-analytics-platform/fractal-web/assets/18033446/aa12da50-b02d-4612-a6a2-bc60ff76d25e">
 
The idea is that each task in the workflow task list shows the processing status as real-time as possible (how real-time is the database for the output ds history? Does it get updated during the run or only at the end of it?)

[...]

PR #363, opened 2023-12-05
Monitoring of job status in workflow page

zonia3000:
## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
+# Unreleased
+
+* Added experimental workflow page with job monitoring (\#363).
+
 # 0.7.0
 
 This release requires fractal-server 1.4.0.
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -38,7 +38,7 @@ export default defineConfig({
 
 	webServer: [
 		{
-			command: './tests/start-test-server.sh 1.4.0a10',
+			command: './tests/start-test-server.sh 1.4.0',
 			port: 8000,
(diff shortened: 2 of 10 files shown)


zonia3000:
The new page is accessible from the path `/projects/{projectId}/workflows/experimental/{workflowId}`

image

It's a preliminary version to check the basic idea. I still have to change the "Run workflow" button to "Stop workflow" when the job starts.

Notice that I've added both input and output dataset dropdowns, as per Figma sketches, but I'm using only the output dataset to retrieve the list of statuses, as described by @tcompa in https://github.com/fractal-analytics-platform/fractal-web/issues/90#issuecomment-1766299615:

> call the new `status` endpoint, with `dataset_id=job.output_dataset_id`

It's not clear to me how to use the combination of input and output datasets in this context.

Moreover I'm never receiving the running status. Jobs remain in submitted status for a while, then go directly to done or failed.

zonia3000:
I've also done some improvements to the workflow task forms, as discussed with @jluethi.

- "Arguments properties" and "Arguments list" texts in accordion headers have been replaced with their related property name
- Accordions containing required properties appears opened by default
- It is now possible to select text inside property description popovers

image

jluethi:
This sounds & looks amazing, looking forward to testing it! 🚀

> I'm using only the output dataset to retrieve the list of statuses, as described by

Yes, only the output dataset has relevant statuses. That's why we're discussing dropping the idea of input datasets :) [see initial issue here, not kept up to date: https://github.com/fractal-analytics-platform/fractal-server/issues/785]
Once we get rid of the concept of an input dataset, we then would only need one dropdown. But to be evaluated & implemented first ;)

Issue #349, opened 2023-11-29
Task collection from whl file: Show in interface that this is possible


jluethi:
I was planning to write an issue to allow this functionality. Turns out, it already works, it just wouldn't be obvious that one could try that from the interface.

----

Using the Fractal task package template (https://github.com/fractal-analytics-platform/fractal-tasks-template), we can now have users create their own Fractal task package. Expecting all those users to deploy those packages to PyPi would be a bit much though. Therefore, it would be very useful to expose the ability to collect tasks based on a whl file in Fractal web.

I currently do this using the CLI fractal-client:
```
fractal task collect /absolute/path/to/gliberal-scMultipleX/dist/scmultiplex-0.4.dev137+g4b46ca5-py3-none-any.whl --package-extras fractal-tasks
```

I tested doing this in Fractal web by using the existing "Collect tasks from a package" option. It basically works, surprisingly!

Basically, I enter the full path to the whl file under package and then click collect.

<img width="1391" alt="Screenshot 2023-11-29 at 14 35 05" src="https://github.com/fractal-analytics-platform/fractal-web/assets/18033446/477ab8c3-e5e2-4d6e-9128-9c9e5b4f023f">

2 limiting factors:
1. It's not clear to me yet what happens to the "Package Version" field in this context? e.g. if I'd set a version different from the package I'm using. To be verified.
2. The "Package" field can be used for that is not obvious (I didn't expect it to work from the web interface, but given the underlying fractal API, I can see why it does work)

tcompa:
This method has been used for testing since the very beginning of fractal-web, and I thought it was known that both package name (for packages on PyPI) or on-disk package path are accepted. This was also meant to support non-public packages being developed internally (as I think some of the fractal-tasks-core-dependent ones were, in the past).

Anyway, let's make it more obvious. Would it be OK to have one of the typical "info" popover button next to "package"? The info message could be something like "The name of a package published on PyPI, or the full path to a wheel file".

---

The version setting is ignored, when using a local wheel file:
![image](https://github.com/fractal-analytics-platform/fractal-web/assets/3862206/ef9056ff-f5dd-4673-b229-990bf8f7a207)

This should probably appear as an API error, however: if the user provides a wheel file, then they cannot also ask for a version.

jluethi:
> Would it be OK to have one of the typical "info" popover button next to "package"? The info message could be something like "The name of a package published on PyPI, or the full path to a wheel file"

[...]

PR #364, opened 2023-12-07
Tasks collection and creation improvements

### LINKED ISSUE #169: Expose `pinned_package_version` in task collection
The `/api/v1/task/collect/pip/` endpoint [receives GET requests with payload like](https://fractal-analytics-platform.github.io/fractal-server/openapi/)
```json
{
  "package": "string",
  "package_version": "string",
  "package_extras": "string",
  "python_version": "string",
  "pinned_package_versions": {
    "additionalProp1": "string",
    "additionalProp2": "string",
    ...
  }
}
```
with [the following optional/required fields](https://github.com/fractal-analytics-platform/fractal-common/blob/main/schemas/task_collection.py):
```python
    package: str
    package_version: Optional[str] = None
    package_extras: Optional[str] = None
    python_version: Optional[str] = None
    pinned_package_versions: Optional[dict[str, str]] = None
```

We should include `pinned_package_versions` as one of the arguments for task-collection requests (in the `/tasks` page). The most relevant use case requires pinning the version of a single package (e.g. `pinned_package_versions = {"torch": "1.12.1"}`), but it would be best to expose a flexible UI that allows any number of entries.

### LINKED ISSUE #314: Make tasks page more compact when many versions of a task are available
Currently, the task page gets quite long and it's hard to keep an overview when many versions of the same task are available. See example here:

Screenshot 2023-10-11 at 16 38 56

It is useful to be able to get the info & edit each task in each version. But the more different versions of tasks we get, the less the simple list view is working to provide an overview.
Can we think of ways to improve this?

One idea: a nested list by task name
I'm also open to other ideas for it :)

### LINKED ISSUE #342: Let user add `args_schema` property when creating a single task
Not a high priority, but this can be handled togheter with #169 to make fractal-web cover all fractal-client features (ref https://fractal-analytics-platform.github.io/fractal-client/reference/fractal/task/#task-new).

### LINKED ISSUE #366: Clear single-task-creation form after success
After the successful creation of a single task, the form remains filled with previous data:
image

I think it'd be best to clear the form, in case of success.

zonia3000:
## Checklist before merging

Closes #169
Closes #342
Closes #366
Closes #314


zonia3000:
On tasks collection it is now possible to add/remove pinned packages versions:

image

On single task creation it is possible to upload an args schema file. The content is validated both to be a valid JSON and a valid JSON schema (using ajv):

image

I've also improved the handling of validation errors received from the server:

image

jluethi:
Great, that looks very useful to have in Fractal web! And very nice validation!

Now that the functionality is getting more complex, we should add a bit more structure to it.

I'd like to tackle the following problems:
1. Make it clear what is required and what is optional. We can go with bold vs. regular. Or we can make the Advanced Options only show when one clicks on expand on them.
Reason: Most people won't need to use pinned package versions, so we don't want to always deploy options that are just needed for some use-cases.

I would prefer a manual separation of arguments, because we support different use-cases. e.g. for "Collect tasks from a package", it supports PyPi packages & whls. Whls don't require the version, I'd encourage specifying it for PyPi packages though.
=> would be good to show Package & version always, have the rest as "advanced options"

For single task, I see Task name, Command & Source as the main options to show. The others are advanced options

2. Add some helper info the the fields. See discussion here on adding a pop-over for the package: https://github.com/fractal-analytics-platform/fractal-web/issues/349
If you can draft some, I can review and suggest text for any missing info box or changes to the text where needed

zonia3000:
I would add a radio button to distinguish between the 2 cases, so that when the user want to collect a wheel package we can hide the version input field and expand the package path input field:

image


image

The advanced options can be put inside an accordion:

image

image

If the layout looks too nested we can simply separate them in this way:

image

Let me know what you prefer.

jluethi:
Oh, I love the PyPi vs. Whl toggle!! And super useful to have the short descriptions below the fields.

While I requested the accordions, I'd say the bottom option actually looks better! Let's go with that!

tcompa:
Thanks Sonia! Slightly unrelated, but while testing this feature I noticed the following minor issues on the tasks page.

1. I think that we should make the separation between "Collect tasks from a package" and "Add a single task" more clear. When we are collecting a task package, the "Add a single task" section appears very similar as yet another option of the same panel (note the white background of the "Add a single task" header):
image

2. Also related: I think that expanding one of the two panels should always make the other one collapse, so that we don't have them both open at the same time (which is currently possible).

3. https://github.com/fractal-analytics-platform/fractal-web/issues/366

Points 1 and 2 make me wonder whether the current organization of the two panels is optimal, or whether we could do something similar to the radio button in https://github.com/fractal-analytics-platform/fractal-web/pull/364#issuecomment-1849656518 also to separate the two task-creation procedures.

---

More specific comments on your proposals:

4. I do like the "PyPI vs local package" radio button, with different forms depending on the choice.

5. It's true that the "extras" argument is optional and advanced, but it's also an argument that we will very often include (this is always true for `fractal-tasks-core`). I suggest we don't hide it behind too many clicks. That is: I suggest we avoid yet another accordion (the one for "Advanced options"), and I'd rather see it as a static form. We are looking at 4/5 arguments, not a thousand ;). OK, I see that now also @jluethi seems to agree (https://github.com/fractal-analytics-platform/fractal-web/pull/364#issuecomment-1851496111).

adrtsc:
Hi, related to this discussion, I was wondering if it makes sense to also allow the user to provide the meta.json file and the args schema version (e.g., pydantic v1.0) when adding a single task? Whenever I create a task on the web interface I still have to add these things to the task with the fractal client via the command line before I can use the task.

Also, it might be nice to have a field for a description when adding single tasks. There is currently a field for "Docs Info" but it can't be edited on the web interface and I think it's not exposed to the fractal client?

tcompa:
Hi there, thanks for chiming in.

> I was wondering if it makes sense to also allow the user to provide the meta.json file and the args schema version (e.g., pydantic v1.0) when adding a single task?

As for the `args_schema_version` attribute: thanks, we actually missed it! This is how it should work: If the user provides an `args_schema` attribute (by uploading a JSON schema file) for a task, then they also must specify the `args_schema_version` attribute, which is a string (typical value: `"pydantic_v1"`).

Note: setting `args_schema` and not setting `args_schema_version` is an invalid behavior, that fractal-web could prevent before an error takes place in `fractal-server`.

tcompa:
> I was wondering if it makes sense to also allow the user to provide the meta.json file

It does make sense, I think. This is how it should work: an additional button "Upload a `meta` file" lets the user upload a JSON file (note: this is not a JSON Schema, just any JSON file), and the corresponding object is used as the `meta` property of the request body.

> Also, it might be nice to have a field for a description when adding single tasks. There is currently a field for "Docs Info" but it can't be edited on the web interface and I think it's not exposed to the fractal client?

Sure. This is how it could work: an additional text prompt, where the user can write a (possibly long-ish, like one or two paragraphs) description of the task, which will be part of the `docs_info` request-body property. This will be eventually rendered in markdown: if there is a simple way of having syntax highlighting, that would be nice (but not if that includes any validation or rendering!).

As @jluethi mentioned above, if we expose all these options we shoukd make sure that "advanced/optional" options are clearly marked as such.

jluethi:
> Points 1 and 2 make me wonder whether the current organization of the two panels is optimal, or whether we could do something similar to the radio button in https://github.com/fractal-analytics-platform/fractal-web/pull/364#issuecomment-1849656518 also to separate the two task-creation procedures.

I like the idea of just having an "Add tasks" section, with 3 options:
PyPi package, Local package, Single task

And with the "Optional Arguments" version, not the expand advanced option seems good.

jluethi:
Thanks @adrtsc for adding these additions from your recent experience with the task addition!
I think it does show the limits of adding single tasks (the amounts of extra info that needs to be provided) and I'd like to steer people towards using the task template package to add those, so it's not adding all of those manually on the web interface each time. But the API is there, so it should also be exposed in Fractal web!


> Sure. This is how it could work: an additional text prompt, where the user can write a (possibly long-ish, like one or two paragraphs) description of the task, which will be part of the docs_info request-body property. This will be eventually rendered in markdown: if there is a simple way of having syntax highlighting, that would be nice (but not if that includes any validation or rendering!).

Let's keep it simple for the time being: There is a text-box that the user enters the optional doc-info. For a better experience, use the task template package (=> PyPi or local package via whl).

----------

To summarize, those are the options to be added:

- [ ] `args_schema_version`: Can we make it default to `pydantic_v1` for the moment, e.g. show that string and the user could change it?
- [ ] "Upload a meta file" for the meta.json
- [ ] `docs_info` text box

zonia3000:
Some more screenshots of the updated UI.

PyPi:

image

Local:

image

Single:

image

Notice that input type and output type are required fields.

I've also added the `docs_link` input field. To reduce the used space the `args_schema_version` field appears only when an args schema file is selected and I've put the documentation fields inside an accordion.

image

tcompa:
Thanks, I think this is an improvement!

Probably the top-menu "Single" label should read something like "Single task".

Second minor comment: does `args_schema_version` only appear upon adding a valid JSON-Schema file? Otherwise I'm confused about why it is missing in one screenshot and present in the following one.

Other than that, we can probably iterate forever about details but I don't see any major missing/unexpected functionality, so I'll refrain from more comments for now ;)

zonia3000:
> Second minor comment: does `args_schema_version` only appear upon adding a valid JSON-Schema file? Otherwise I'm 
confused about why it is missing in one screenshot and present in the following one.

The `args_schema_version` field appears only when an args schema file is selected. I did that to reduce the number of fields always visible to the user. The file is validated when the user clicks on the "Create" button though.

jluethi:
This looks great!!

> Probably the top-menu "Single" label should read something like "Single task".

Agreed, "Single task" is good.

I'd also say this look like a very good improvement now. I'll run some local tests (hopefully tomorrow) to provide a final round of feedback and see if any behaviors are not as expected. We can always improve on it more later of course :)

jluethi:
Running task collection for the fractal-tasks-core package locally has quite a long delay between clicking "Collect" and the task collection section appearing below. I initially thought it wasn't working, but it just took a few seconds.
I remember that being a faster interaction. Any idea what could slow it down? It did appear faster when I then collected a second package with the section already present on the page

jluethi:
Besides this performance question, all looks good to me! Great improvement to the task collection interface and awesome that we reach feature parity here with the CLI client! And the new task list is much easier to keep an overview on!

And I can't always reproduce this performance issue exactly with e.g. different versions of fractal-tasks-core or with starting a fresh server and collecting again. Is it dependent on a response from pypi? I noticed that when I request a package that isn't available, I get a 422 error server side. This error is shown above the task collection, not as a failed task collection. Does the server do validation of the request before confirming it?

The thing I want to avoid: Users pressing task collection 3x, because "nothing is happening". If we want to keep the PyPI validation of package names (I can see that this is useful!), we can also address this by e.g. making the Collect button not clickable until we get a response from the server.

zonia3000:
In some occasions I experienced slow responses for that endpoint too. We are checking with pip if the package exists and I think that slow responses might depend on PyPi repository server being temporary overloaded.
I can disable the button and add a small spinner inside it while the API call is in progress to improve the feedback to the user and avoid multiple clicks.

jluethi:
Great, let’s go with that then! :)

jluethi:
Ah, one last thing: Could we make the statuses on the task collection auto-update easily? I like the auto-updates on the job page, would be neat to have them here as well :)

jluethi:
After that, I'll stop adding requests to this PR and we can merge it & release Fractal web 0.7.1 :)

zonia3000:
I implemented the task collection auto-update. Would you like also that the tasks table is refreshed when a task moves from `installing` to `ok` status?

jluethi:
> I implemented the task collection auto-update. Would you like also that the tasks table is refreshed when a task moves from installing to ok status?

That would be very nice!

jluethi:
Just tested the spinner & the auto-updating, both are very nice!
If you also add the refreshing of the task list once a collection is done, let's merge that and release as 0.7.1 :)

zonia3000:
I think we can also remove the refresh buttons from the tables. Do you agree?

jluethi:
If it auto-refreshes, then yes! :)

zonia3000:
Done :)

jluethi:
Tested it. It's awesome! What a nice set of improvements to the task collection page!

Let's update the Changelog to represent the additional changes, merge and make this the 0.7.1 Fractal web release! :)

PR #365, opened 2023-12-11
Supported editing of cache_dir from user profile page

### LINKED ISSUE #348: Let user modify their own attributes
_[requires fractal-server >= 1.4.0a8]_

The relevant endpoint is `PATCH /auth/current-user/`, which only accepts a payload like
```json
{
  "cache_dir": "/some/absolute/path"
}
```

Notice that the API call will fail for:
* `{"cache_dir": null}` (in the future we may change this, but for the moment it raises an error)
* `{"cache_dir": ""}`
* `{"cache_dir": "   "}`
* `{"cache_dir": "non/absolute/path"}`
* `{"unexpected": "attribute"}`

---

I see at least two options:
1. In the user's profile page, we make the `cache_dir` text field editable - with something like a "save" button next to it
2. We have two different pages for viewing and editing the profile

Note that for the moment there will be a single editable field but this could change in the future, either because we include the password (which is currently not allowed - ref https://github.com/fractal-analytics-platform/fractal-server/issues/1018) or because to let the user modify their slurm accounts (ref https://github.com/fractal-analytics-platform/fractal-server/issues/956 and #334).

zonia3000:
## Checklist before merging

Closes #348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Unreleased
 
+* Supported editing of cache_dir from user profile page (\#365).
 * Added experimental workflow page with job monitoring (\#363).
 
 # 0.7.0
--- a/src/routes/profile/+page.svelte
+++ b/src/routes/profile/+page.svelte
@@ -1,50 +1,140 @@
 <script>
 	import { page } from '$app/stores';
+	import { displayStandardErrorAlert, getValidationMessagesMap } from '$lib/common/errors';
 	import BooleanIcon from '$lib/components/common/BooleanIcon.svelte';
 
 	/** @type {import('$lib/types').User} */
 	$: user = $page.data.userInfo;
+
+	/** @type {import('$lib/components/common/StandardErrorAlert.svelte').default|undefined} */
+	let errorAlert = undefined;
+
+	let editCacheDir = false;
+	let cacheDir = '';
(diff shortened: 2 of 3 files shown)


zonia3000:
I added an edit button next to the cache dir field, that is the only field we can edit at the moment:

image

The button shows an input field that is hided on save/discard action:

image

Cache dir validation error are displayed near the field:

image

tcompa (approved):

tcompa:
> Quick server question: If we get a validation error that Cache Dir can't be empty (makes some sense), why could the user be created without a Cache dir?

The column is optional (AKA nullable), which means its value can be missing (AKA null, AKA `None`).
The error appears because we are trying to set it to an empty string, which is not accepted (as a non-absolute path would not be accepted).

This means that in principle the current form is missing an option, namely the possibility for a user to remove their own `cache_dir`.

However we currently switched off this option, in `fractal-server`, because of its real-life consequences. When running on SLURM, the `cache_dir` _is_ required, and jobs will fail if it's not there. We (with @ychiucco) thought it's better to prevent the user from entering this branch. What is doable for the user is only to edit their `cache_dir`, but it should still be a valid absolute path.

jluethi:
Ok, makes sense, let's keep that for the moment then :)

Issue #161, opened 2023-06-09
Automatically test fractal-tasks-core task-argument JSON Schemas


tcompa:
As part of the new vitest tests, we should also loop over the latest fractal-tasks-core schemas (WARNING: they are evolving a lot in this period, thus some issues may be unrelated to fractal-web); note that this first requires #155.

This is how one could fetch these schemas in a Python script
```python
import requests

url = "https://raw.githubusercontent.com/fractal-analytics-platform/fractal-tasks-core/main/fractal_tasks_core/__FRACTAL_MANIFEST__.json"
manifest = requests.get(url).json()
schemas = [task["args_schema"] for task in manifest["task_list"]]
```
and the same logic can be included in a test here in fractal-web.

zonia3000:
After discussing with @tcompa we decided to close this issue.

In the past it could be useful to test the rendering of the form because the rendering could fail if the component encountered an unsupported type. However, at the moment the rendering of the form can fail only if an invalid JSON schema is provided and fractal-tasks-core JSON Schemas validity is already tested in the back-end (see https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/tests/tasks/test_valid_args_schemas.py#L126-L142). So, merely testing the rendering of the form is not very useful in this case. A more complete test that verifies the expected rendering of each field would be more useful, but that should be kept aligned with the evolving fractal-tasks-core versions and so requires a lot of effort.

PR #393, opened 2024-01-17
Worflow task form improvements

### LINKED ISSUE #177: Support for `enum` in JSON Schema
Example
```python
from typing import Literal
from pydantic import BaseModel
import json

class Args(BaseModel):
    x: Literal["option1", "option2"]

print(json.dumps(Args.schema(), indent=2))
```
produces
```json
{
  "title": "Args",
  "type": "object",
  "properties": {
    "x": {
      "title": "X",
      "enum": [
        "option1",
        "option2"
      ],
      "type": "string"
    }
  },
  "required": [
    "x"
  ]
}
```

I guess that we don't currently support a property with `enum`, in the UI.

The ideal feature would be that for `x` we offer a drop-down menu to choose out of `option1` and `option2`, but we can perhaps postpone its implementation. The ajv validation will correctly interpret this schema, and then it will fail if we set `x` to `option3`.

### LINKED ISSUE #183: Integrate additional JSON Schema constraints (e.g. min, max, max length)
Be aware that some values defined by the schema could have restrictions as specified by the properties in the title.

It is currently not required by the client to highlight those restrictions/requirements as a mean of UI state validations or complex UI interactions that could happen with the user.

The client as a whole is already compliant with those restrictions thanks to the AJV validations.

### LINKED ISSUE #193: Reorder items of an argument which is an array
Low priority

### LINKED ISSUE #202: Review required/optional field behavior
It is currently supported (we highlight the required parameters), but only for first-level parameters, not for nested ones. This is a minor issue, but should be fixed.

This requires:
- [ ] Display some "highlighting" in the UI
- [ ] Review that behavior is the expected one

zonia3000:
## Checklist before merging

Closes #177
Closes #183
Closes #193
Closes #202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     * supported JSON Schema `minimum`, `maximum`, `exclusiveMinimum` and `exclusiveMaximum` for numeric types;
     * supported JSON Schema `minItems` and `maxItems` for array type;
     * supported reordering of array fields;
+    * set bold font on nested required properties;
     * improved field validation error messages.
 
 # 0.8.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,15 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
-# 0.8.1
+# Unreleased
 
+* Improvements on worflow task form (\#393):
+    * supported JSON Schema `enum` type;
+    * supported JSON Schema `minimum`, `maximum`, `exclusiveMinimum` and `exclusiveMaximum` for numeric types;
+    * supported JSON Schema `minItems` and `maxItems` for array type;
+    * supported reordering of array fields;
+    * improved field validation error messages.
+
+# 0.8.1
 
 Note: with this release, `PUBLIC_FRACTAL_ADMIN_SUPPORT_EMAIL` becomes a required
 environment variable, and `npm run build` fails if it is missing. If you do not
--- a/__tests__/JSchema.test.js
+++ b/__tests__/JSchema.test.js
@@ -1,0 +1,103 @@
+import { describe, it, expect } from 'vitest';
+import { render } from '@testing-library/svelte';
+
+import JSchema from '../src/lib/components/common/jschema/JSchema.svelte';
+
+describe('JSchema', () => {
+	it('NumberProperty with title', async () => {
+		const result = renderSchemaWithSingleProperty({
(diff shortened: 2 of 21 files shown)


tcompa:
Since we are already dealing with tests for the JSON schema component, would it make sense to also address https://github.com/fractal-analytics-platform/fractal-web/issues/161 here? It is about fetching (from github) a list of JSON schemas, and checking that they can all be transformed into UI components with no errors raised.

zonia3000:
JSON Schema `enum` type is now supported as a dropdown:

image

For numeric types we now support `minimum`, `maximum`, `exclusiveMinimum` and `exclusiveMaximum`. A user friendly error message appears if the user inserts a number that is greater than or lower than the expected range. Moreover, the `min` and `max` attribute are set to the HTML input, so, if the user increments or decrements the number using the small buttons inside the input fields, the value will never be outside the range.

image

For array types `minItems` and `maxItems` are now supported. In the following example the `minItems` value is set to `2` and the form shows by default the minimum required number of fields (only if the field is also in the `required` list). It is not possible to remove the items when they are less than the `minItems` value. In this situation the Remove button empties the field but doesn't remove it.

image

When the `maxItems` number is reached it is not possible to add more items anymore:

image

Moreover, array items can now be reordered.

jluethi:
Oh that's awesome! @tcompa Do we also have tuples (or fixed-length lists) on our list somewhere? Would be a useful addition to become more general!

tcompa:
> @tcompa Do we also have tuples (or fixed-length lists) on our list somewhere?

I think that a Python tuple simply gets converted into a JSON-schema array with min/max, and then it is fully covered by:

> For array types minItems and maxItems are now supported. 

---

Note, however, that I'm currently not encouraging the use of tuples for task arguments - because of https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/564. That's why you won't currently find `minItems` or `maxItems` in fractal-taks-core manifest - although we eventually will need to support it (likely after the Pydantic v2 switch, since it targets Draft 2020-12 which is the one where the breaking change appeared).

tcompa:
To be more clear: out of the new features, I think for the moment we're only using the new enum feature (in the napari-workflows task). Still, it's great to be supporting more JSON-schema features (thanks @zonia3000!).

jluethi:
For sure! And with Enums supported, we could revisit the option to e.g. use them for the selection of networks in the Cellpose task.

zonia3000:
Required properties of nested objects are now set to bold:

image

PR #410, opened 2024-02-09
Implemented import and export of workflow task arguments

### LINKED ISSUE #245: Enable importing/exporting single-WorkflowTask arguments as JSON
### Feature
On top of having two alternative ways of setting task parameters (either via a JSON Schema form, or via a free form in case there is no available schema), we should expose the option to directly set all parameters from an existing JSON file (or otherwise from an existing JSON string to be copy-pasted into a text box - TBD which one is more useful).
Also the reverse should be possible ("view task parameters as a JSON string", or "export task parameters to a JSON file").

### Perspective
These two features, together, would enable a first straightforward way to transfer the task parameters from an "experimental" workflow to a "production" one. Note that, depending on how we will structure this kind of use flow, we may combine the export-and-then-import feature into a single "copy-these-parameters-to-a-task-of-another-workflow" feature/button. But we the straightforward implementation is useful to play around, and it could also always keep a relevant use case ("can you please send me by email the full JSON of those very cumbersome-to-set create-ome-zarr parameters?").

### LINKED ISSUE #408: Task-addition modal shows empty list, the first time it's opened
### Observed behavior

When adding a task to an empty workflow for the first time, the list appears as empty and only showing "search".

image

After a few clicks to expand/compress the list, it appears as pre-filled with available tasks

image

### Expected behavior (IMO)

The list should appear as pre-populated every time, including the very first time.

zonia3000:
## Checklist before merging

Closes #245
Closes #408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+* Added spinner on workflow task modal when tasks list is loading (\#410).
+* Implemented import and export of workflow task arguments (\#410).
 * Improved sorting of users in dropdown of the admin jobs page (\#402).
 * Fixed bug in retrieval of job log from the admin jobs page (\#402).
 * Highlighted relevant part of the error message in workflow job log modal (\#402).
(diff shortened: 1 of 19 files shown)


tcompa:
I just tested this, and it works as expected.

A minor comment: canwe change the `args.json` filename into something task-specific? E.g. I would use a slug version of `workflowtask.task.name`, as in `args_{task_name_slug}.json`.

tcompa (approved):

PR #424, opened 2024-03-04
Added search functionality on jobs table filters

### LINKED ISSUE #423: On job page, add search functionality for Project, workflow, input & output dataset dropdowns
I just saw a user navigating these dropdowns with some 30 entries or so and a search box would really have helped. Are there downsides to adding a search box to these dropdowns? I guess we're already loading the list of all datasets etc. to generate the dropdown, right?

zonia3000:
## Checklist before merging

Closes #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
+# Unreleased
+
+* Added search functionality on jobs table filters (\#424).
+
 # 0.10.1
 
 * Fixed bug in workflow task meta properties update (\#422).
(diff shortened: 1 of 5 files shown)


tcompa:
I notice something slightly unexpected, but perhaps this was not even introduced within this PR.

I have this list
image

I search for workflow 222 in the search box, and then click it - so that I am now filtering by workflow 222
image

I click the X next to the search box (where I had searched for workflow 222), and there is no active filter shown in the workflow column. But the filter is actually being applied. This doesn't look fully consistent to me. 
image

Note that I can still disable the filter by clicking on the yellow "clear filters" button. But I would expect that if the filter is active then this information should be explicitly displayed.

zonia3000:
@tcompa I think it should be fixed now, however I wasn't able to reproduce it using the X button, but only clicking on the selected element again in order to unset the selection.

tcompa (approved):

PR #434, opened 2024-04-05
API V2 Support

### LINKED ISSUE #426: Image-list page (first version)
In V2 dataset has a `images` column.

# POST /api/v2/project/111/dataset/222/images/query/, basic usage

Basic behavior (if there is no request body): return full list of images of given dataset

Query-parameters:
* `page_size`, default to null (server-side). Null value means "give me the whole list"
* `page`, default to 1 (server-side)

```
{
 "total_count": 4353,  # always present
 "page_size": 50  # the one required by the client, or `total_count`
 "page": 1,  # always present
"attributes": ["key", "another_key"],  # this is the full list of attributes, not the one for the filtered list
"images": [
      {"path": "/tmp_zarr_dir/plate.zarr/B/03/0", "attributes": {"key": true}},
      {"path": "/tmp_zarr_dir/plate.zarr/B/03/1", "attributes": {"another_key": 1234}},
  ],
}
```

Relevant assumptions:
* All images must have a `path` property
* `path` values are unique across the list of images
* all `path` values must be relative to `dataset.zarr_dir`

Client-side display:
* display a table, with given columns (provided in response.attributes), and an "Options" column (for the moment with only the "delete" button -> pointing to the DELETE enpdoint described below)
* extract `zarr_dir` from `GET /api/v2/project/111/dataset/222/` (e.g. `zarr_dir=/tmp_zarr_dir`) and left-strip it from the paths (that is, display `plate.zarr/B/03/0"`)
* Also display `zarr_dir`, before the table
* Also display `total_count` (either in the pagination footer, or somewhere above the table)

# POST /api/v2/project/111/dataset/222/images/query/, advanced usage

Also include a payload to filter the image list.
```
{
    "path": "/tmp/asdas",  # optional, must be a string
    "attributes": {"key": false},  # optional, can only be number, boolean, or string
}
```
Both path and attributes are optional, and they can co-exist.
If path is set, then either 0 or 1 images are return.

The form could look like this one:
image

The key names can be chosen out of the known "attributes" property (plus the "path" one, which is always possible and always the first one).

# DELETE /api/v2/project/111/dataset/222/images/?path=encoded_path

With a confirmation dialog.

### LINKED ISSUE #427: Dataset create/view/edit (first version)
V1 attributes that are not present in v2:
* `input_type` and `output_type`
* `resource_list`
* `meta`

Create dataset (POST /api/v2/project/1/dataset/)
* `zarr_dir` (required, string, absolute path)
* `filters`: dict[str, Union[str, float, int, bool], with arbitrary keys. Example of dataset.filters: `{"key": 1, "another_key": true, ..}`. User should be able to add new keys. Property values are scalar, i.e., they are not arrays/objects.

View dataset (GET /api/v2/project/1/dataset/1/)
* `name` - same as V1
* `zarr_dir`
* `read_only` - same as V1
* `filters`:
* `history` - same as V1

Edit dataset (PATCH /api/v2/project/1/dataset/1/)
* `name` - same as v1
* `zarr_dir`
* `read_only` - same as V1
* `filters`: if request body has `filters` property, this fully replaces existing value. Therefore it's possible to remove a given property.

cc @ychiucco - let's also clean up the schemas

### LINKED ISSUE #429: Exclude V1/V2 reserved arguments from JSON-schema-based form
(just a quick note, since it came up during testing)

In V1 we have a block like
```js
export function stripSchemaProperties(schema) {
	const ignoreProperties = ['input_paths', 'output_path', 'metadata', 'component'];
	Object.keys(schema.properties).forEach((k) => {
		if (ignoreProperties.includes(k)) {
			delete schema.properties[k];
		}
	});
	if (schema.required) {
		schema.required = schema.required.filter((k) => !ignoreProperties.includes(k));
	}
}
```

For a V2 WorkflowTask this changes depending on the `workflowtask.is_legacy_task` boolean flag.
When we are using a legacy task (`workflowtask.is_legacy_task=true`), then the `ignoreProperties` array is the same as in V1.
When we are using a current task (`workflowtask.is_legacy_task=false`), then the `ignoreProperties` array is `['path', 'paths', 'init_args', 'zarr_dir']`.

(there may be minor changes in the actual values in the list, but for sure it will differ from what we have in V1)

### LINKED ISSUE #430: (minor) Replace "flags" with "types"
I spotted a minor glitch in the dataset modal, where "flag" should read "type":
image

### LINKED ISSUE #431: Import/export WorfklowTask V2 arguments
How do import-export args work for the three types of V2 tasks (compound, parallel, non_parallel)?

Naive approach: both parallel and non-parallel args are exported, in a structured object.
1. For a compound task: `{"args_parallel": {...}, "args_non_parallel": {...}}`.
2. For a parallel task: `{"args_parallel": {...}, "args_non_parallel": null}`, or even  `{"args_parallel": {...}}`.
3. For a non-parallel task, similar to a parallel task.
4. Upon import, we combine the object content with the task type, and fill only the arguments that are relevant (e.g. only the parallel ones).

(@zonia3000, let me know if this looks reasonable)

### LINKED ISSUE #432: Handle version updates for V2 tasks - TBD
How should the potential version updates of tasks V2 be handled? Here is the first plan.

Can we assume that different versions of the same task have the same type (compute, parallel, non_parallel)?
This is currently not guaranteed (backend-side), and then the frontend should not include NEW_TASK as a possible update of OLD_TASK if their types are different.

zonia3000:
## Checklist before merging

Closes #426
Closes #427
Closes #429
Closes #430
Closes #431
Closes #432


tcompa:
In single-task creation of a non-parallel task, I see the following error if I include the non-parallel meta JSON file:

image

Maybe a leftover `meta` property in the payload, which should rather be `meta_non_parallel`?

tcompa:
In the single-task addition, let's name the task types as "Non parallel", "Parallel", "Compound" (without "Standalone"):

image

tcompa:
In the tasks v2 page, how are tasks sorted? It seems to be neither by ID nor by name.

image

I propose we proceed as for V1 (https://github.com/fractal-analytics-platform/fractal-web/issues/149#issue-1713800874):

> Change order of tasks in the /tasks page. Sorting should be according to task.owner (first the ones with null, then the others), then task.name

tcompa:
Job submission endpoint path was changed from
```
/api/v2/project/1/workflow/1/apply/?dataset_id=1
```
to
```
/api/v2/project/1/job/submit?dataset_id=1&workflow_id=1
``

**Comment by tcompa** [comment]
I forgot that the admin-area pages also require the `/admin/v2/project/` endpoint, which was recently removed from v2 - and this is what I see now:
```
INFO:     127.0.0.1:49048 - "GET /admin/v2/project/ HTTP/1.1" 404 Not Found
```

We will re-enable it in fractal-server asap.

zonia3000:
> In the tasks v2 page, how are tasks sorted? It seems to be neither by ID nor by name.

The sorting function was the same as v1. The issue was that by default sorting is case sensitive and uppercase letters are before the lowercase ones, so "MIP" is before "apply" in case sensitive comparison. Made case insensitive both for v1 ans v2.

tcompa:
Within the single-task-creation page, could we forbid providing the same input-type key twice?
image
(and the same holds for output-type)

tcompa:
A minor glitch while handing an edge case in the single-task-creation page.
1. Fill the form for a e.g. parallel task, and add a meta file.
2. Remove the file from disk
3. Click "Create"

Nothing happens, but the browser console has this error
image

tcompa:
Again in the single-task-creation page, the keys for the input/output types should never be empty strings.
This is something we will also validate backend-side (ref https://github.com/fractal-analytics-platform/fractal-server/issues/1354), but it doesn't hurt to make it impossible already within the frontend.

tcompa:
In the task-info modal, let's include also `meta_parallel` and `meta_non_parallel` - which are currently not shown:
image

tcompa:
In the dataset/images page, we will frequently hit very long paths. Let's review whether on or both of the behaviors below can be improved (e.g. by wrapping?)

image
image

tcompa (approved):

zonia3000:
The error in meta when the browser can't read the selected file:

image

The wrapped text in modal:

image

PR #450, opened 2024-04-12
Tasks V1/V2 compatibility

### LINKED ISSUE #330: Prepare tasks packages for testing
Requirements:
1. Super lightweight (pydantic only)
2. They can run
3. They can fail, given a certain input parameter
4. Different versions, possibly with changes in JSON Schemas

There exist previous versions (https://github.com/fractal-analytics-platform/fractal-demos/tree/main/examples/99_stress_test/fractal-tasks-stresstest), but it'd be better to use the new fractal-tasks-template.

### LINKED ISSUE #433: Placeholder: support using V1 tasks within V2 WorkflowTask
(ref https://github.com/fractal-analytics-platform/fractal-server/issues/1328)

### LINKED ISSUE #438: Add v1-task forward-compatibility section to v2 admin area
As discussed elsewhere, an admin should be able to make `PATCH /admin/v2/task-v1/{task_id}/?is_v2_compatible=true` or `PATCH /admin/v2/task-v1/{task_id}/?is_v2_compatible=false` requests.

The first option we came up with is to add a new section into the v2 admin area, named like "Tasks V1/V2 compatibility". This page should then show a simplified list of the V1 tasks (simplified -> it does not expose the edit/delete options), where for each task the admin can set the `is_v2_compatible` flag to true or false.

The other option we explored was to modify the current V1-tasks page, and add a hidden feature (which is shown for admins only) where this same kind of edits can be done. We opted for the separate admin-page to keep this specific feature confined in its own place (where it can also be removed very easily, in the future) and to avoid including complexity in a page which is visible to standard users.

### LINKED ISSUE #448: Remove all read-only from V2 project/dataset components
Ref https://github.com/fractal-analytics-platform/fractal-server/issues/1384

zonia3000:
## Checklist before merging

Closes #438
Closes #433
Closes #448
Closes #330


zonia3000:
The new Task V1/V2 compatibility page:

image

Switcher for V1/V2 tasks in workflow page task selector:

image

Legacy tasks inserted in V2 workflow:

image

tcompa:
> The new Task V1/V2 compatibility page:

Looks good to me!
How are we displaying different versions of the same task? Are they still all listed (even if they are not grouped any more)?

> Switcher for V1/V2 tasks in workflow page task selector:

I think we should avoid V1 and V2. Maybe "current tasks" and "legacy tasks"?

> Legacy tasks inserted in V2 workflow:

Same as above, what about "legacy" instead of "v1"?

tcompa (approved):

zonia3000:
With multiple versions the old versions are inside collapsed rows, as in the main task list table:

image

Used current/legacy:

image

jluethi:
+1 on using legacy wording.

I think Tommaso's question was for the adding task modal. There, the version selector usually just shows up when a task is selected, I assume it's still the same :)

tcompa:
Task V1 collection:
image

Task V2 collection:
image

Admin V2 sections:
image

Setting V1 tasks as V2-compatible (with single versions):
image

Setting V1 tasks as V2-compatible (with multiple versions available):
image

Minor question: before I expand the "+" button, I can already mark act on the true/false switch, which modifies the latest-version task. I'm not sure this is the most intuitive behavior (at first I thought it would have modified the whole block of tasks, for all versions), but TBH this is an admin-only page and we don't need to invest in polishing too much.

Adding a V2 task into a V2 workflow
image

Adding a V1 task into a V2 workflow (when no V1 task is V2-compatible):
image


Adding a V1 task into a V2 workflow (when there is a single version of the only V2-compatible V1 task):
image

Adding a V1 task into a V2 workflow (when there are two versions of the only V2-compatible V1 task):
image

tcompa:
All works as expected, thanks!
Let's open other issues for polishing this new feature.

PR #465, opened 2024-04-19
Continue restart workflow

### LINKED ISSUE #440: Support different submission scenarios (start/continue/restart)
The V2 workflows seem to be very optimized for continuing workflows from where one has stopped. But another typical behavior is that a user reruns the whole workflow from the beginning.

That gets confusing, because there can already be an existing image list and existing filters that are set. For example, if a user runs a workflow that goes Convert OME-Zarr, MIP.
Then they want to rerun it, it fails. Because the is_3D filter was set to False and rerunning Convert OME-Zarr does not change the filter.

I haven't fully figured out what we'll need to handle this better. Here are a few points:

- We'll need to expose what filters are already active when a workflow gets submitted.
- I think we may need different concepts for rerunning a workflow from the beginning vs. continuing an existing workflow

Rerunning a workflow: 
- What happens with the existing image list?
- What happens with the existing filters?

zonia3000:
## Checklist before merging

Closes #440

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
     * moved editing of dataset inside dataset page (\#452);
     * supported editing of single dataset images (\#457);
     * used switches to represent boolean flags (\#457);
+    * implemented continue/restart workflow (\#465);
 
 # 0.10.2
 
--- a/__tests__/job_utilities.test.js
+++ b/__tests__/job_utilities.test.js
@@ -1,5 +1,9 @@
 import { it, expect } from 'vitest';
-import { extractJobErrorParts, extractRelevantJobError } from '$lib/common/job_utilities.js';
+import {
+	extractJobErrorParts,
+	extractRelevantJobError,
+	generateNewUniqueDatasetName
+} from '$lib/common/job_utilities.js';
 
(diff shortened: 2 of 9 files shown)


tcompa:
image

When I set the new dataset name to be identical to the old one, I'm expecting something to fail backend side, but apparently this is not the case since I can click "continue" with no problem. I checked in the dataset page, and I can create N datasets with the same with no problem.

I don't know, right now, whether this is the expected behavior server side. But I would start by preventing this from happening when restarting a workflow, if "Replace existing dataset" is unchecked. Otherwise we end up with two datasets with the same name.

tcompa (approved):

tcompa:
> Just a curiosity: how are the suffixes created? When the original dataset name is ds-cardiac-test-partial-2_1, I was naively expecting the proposed new name to be ds-cardiac-test-partial-2_2 - but I found ds-cardiac-test-partial-2_1_1. How does this work?

Now clarified: if the dataset name does not end with `_{integer}`, then we add `_1`. If it does end with `_{integer}`, then we propose to use `_{integer+1}`.

zonia3000:
The message appearing when attempting to reuse the same name for a new dataset:

image

tcompa:
Minor: "different than" -> "different from"

PR #466, opened 2024-04-23
Workflow execution modal improvements

### LINKED ISSUE #442: Show current filters during workflow submission & allow changes
The workflow submission form should show what filters are currently set. 

As a user, I can see what filters would be used for a workflow submission.
(potentially I can also see what images would then be used for processing)

As a user, I can change what filters will be set during workflow submission.

I haven't figured out the UI aspects for this. Is this currently feasible from a backend perspective @tcompa ?

### LINKED ISSUE #464: How to set default starting task when continuing a workflow execution
Say we have a workflow task list that includes 4 tasks. Each one of them has a status which can be
* done
* failed
* submitted
* `-` (not present in the API response)

Let's go through a few scenarios, and determine what is the expected default for where the workflow should restart from. If this logic is a function of a (wftask_id, wftask_status) list, then I think this feature should be implemented directly in the frontend.

Of course it relies on what are the specification for the get-status endpoint, which easily hit edge cases when the task list is modified (see https://github.com/fractal-analytics-platform/fractal-server/issues/1083#issuecomment-2063898906). Those edge cases often propagate into the current issue.

### A

I deleted all tasks that already ran:

1 -
2 -
3 -
4 -

Proposed first task: 1
(but this likely a case where we do not offer a "continue" feature, then it's not very relevant - see #440)

### B

I only ran two tasks:

1 done
2 done
3 -
4 -

Proposed first task: 3

### C

I ran three tasks, and the third one failed:

1 done
2 done
3 failed
4 -

Proposed first task: 3

### D

By playing with the task list, the task that failed is not the last one:

1 done
2 failed
3 done
4 -

Proposed first task: 2

### E

By playing with the task list, I create a "hole":

1 done
2 -
3 done
4 -

Proposed first task: 2

### F

By playing with the task list, I create a "hole" and then a failed task:

1 done
2 -
3 failed
4 -

Proposed first task: 2

### G

1 done
2 done
3 done
4 done

Proposed first task: 4 or none?

### H

1 submitted
2 failed
3 -
4 -

Proposed first task: N/A, because we cannot re-submit while something is running.

### I

Anything else? I listed the first ones coming to mind.

---

Note that the basic logic I used, above, is that `-` and `failed` are treated in the same way. I cannot say yet whether this always holds.

---

Independently on how far we go in covering edge cases, we could also make this feature "soft" as in "if the scenario does not fall into a list of supported cases, then do not propose any specific first task but highlight the first-taks-selection element".

zonia3000:
## Checklist before merging

Closes #464
Closes #442

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
     * supported editing of single dataset images (\#457);
     * used switches to represent boolean flags (\#457);
     * implemented continue/restart workflow (\#465);
+    * set default first task when continuing a workflow (\#466);
 
 # 0.10.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
     * used switches to represent boolean flags (\#457);
     * implemented continue/restart workflow (\#465);
     * set default first task when continuing a workflow (\#466);
+    * displayed applied filters in workflow execution modal (\#466);
 
 # 0.10.2
 
--- a/__tests__/job_utilities.test.js
+++ b/__tests__/job_utilities.test.js
@@ -2,7 +2,8 @@ import { it, expect } from 'vitest';
 import {
 	extractJobErrorParts,
 	extractRelevantJobError,
-	generateNewUniqueDatasetName
+	generateNewUniqueDatasetName,
+	getFirstTaskIndexForContinuingWorkflow
 } from '$lib/common/job_utilities.js';
 
 const completeTracebackError = `TASK ERROR:Task id: 15 (Create OME-Zarr structure), e.workflow_task_order=0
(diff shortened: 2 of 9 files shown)


tcompa:
Something looks confusing.

The dataset has filters:
image

The submission modal does not show the dataset filters:
image

tcompa:
> Something looks confusing.

We now understand this.
What was not obvious is that `filters` (like `images` and `history`) can be modified during workflow execution, and then they should be loaded afresh when needed.

tcompa:
As of https://github.com/fractal-analytics-platform/fractal-web/pull/466/commits/e9ff01c718d47ed56504586a623ac55dff37f245, the issue I spotted is now fixed.

tcompa:
Overall, this works as expected from my side.

tcompa (approved):

PR #467, opened 2024-04-24
Implemented import and export of datasets

### LINKED ISSUE #445: Can we make the workflow page default to an appropriate dataset?
Defaulting the workflow page to not showing a dataset is not that intuitive. Can we handle the default smartly, e.g. the last selected dataset is shown as default again?
We don't have that information server-side, but maybe we can cache it client-side?

Alternatively, do we have a "last edited" timestamp for dataset? That would probably be the dataset a user often wants to work with. If that's not available, most recently created may be good.

Best case: Default to last selected, if nothing has been selected in a given session yet, default to the last edited/last created dataset

### LINKED ISSUE #451: Should we add an help message to the insert-legacy-task feature?
As of #450, this is the task-insertion modal:

image

Something that I guess could happen is that a user expects to find their V1 tasks in the list, but they are not. W could add a small help message (possibly requiring a click on a "help" or "info" button/link), that says something like "Only certain legacy tasks can be run within the new Fractal (see link): if you want to add one of yours, ask an admin".

Opinions?

### LINKED ISSUE #456: Add import/export-dataset features
# Export

Endpoint:
```
GET /api/v2/project/{project_id}/dataset/{dataset_id}/export/
```

Response-body model:
image

# Import

Endpoint:
```
POST /api/v2/project/{project_id}/dataset/import/
```

Request-body model:
image

Response-body model: same as the standard POST/GET dataset endpoints.

### LINKED ISSUE #468: Continue Workflow Dialogue: Highlight First task when no default is present
The Continue Workflow modal is currently not very obvious if there is no default first task selected. The run button is greyed out, because no first task is selected and that's required. We should highlight the missing field better (bold, red or something).

The reason no default for the First Task was chosen here is that the workflow had finished, so there wasn't an obvious default (=> makes sense!). But we should highlight it to the user where input is required.

Screenshot 2024-04-24 at 14 56 33

Just for reference, when only a partial workflow was run, the modal correctly defaults to the next task, and then all the current behavior makes sense and we don't want to highlight the First task selection more.
Screenshot 2024-04-24 at 15 02 43

### LINKED ISSUE #469: In Workflow Task page, rename titles for non-parallel & parallel
Currently, the titles for the parameters are shown as `Args non parallel` and `Args parallel`. This is not very intuitive for users. Let's rename them to:

1. Initialisation Parameters
2. Compute Parameters

Screenshot 2024-04-24 at 16 12 33

### LINKED ISSUE #470: Why are `Args parallel` shown when there are no parallel args to be set?
The newest version of the `Convert Cellvoyager Multiplexing to OME-Zarr` task (not in a prerelease yet) has no more parallel-args. The web interface still shows a section for it though. Can we not show the init / parallel Args sections when there are no arguments?

Screenshot 2024-04-24 at 16 15 55

There are arguments to that task, but none are user-exposed in the web interface.

For reference, it has a schema like this:
Args Schema non parallel
```
{
  "title": "CellvoyagerToOmeZarrInitMultiplex",
  "type": "object",
  "properties": {
    "acquisitions": {
      "title": "Acquisitions",
      "type": "object",
      "additionalProperties": {
        "$ref": "#/definitions/MultiplexingAcquisition"
      },
      "description": "dictionary of acquisitions. Each key is the acquisition identifier (normally 0, 1, 2, 3 etc.). Each item defines the acquisition by providing the image_dir and the allowed_channels."
    },
    "image_glob_patterns": {
      "title": "Image Glob Patterns",
      "type": "array",
      "items": {
        "type": "string"
      },
      "description": "If specified, only parse images with filenames that match with all these patterns. Patterns must be defined as in https://docs.python.org/3/library/fnmatch.html, Example: `image_glob_pattern=[\"*_B03_*\"]` => only process well B03 `image_glob_pattern=[\"*_C09_*\", \"*F016*\", \"*Z[0-5][0-9]C*\"]` => only process well C09, field of view 16 and Z planes 0-59."
    },
    "num_levels": {
      "title": "Num Levels",
      "default": 5,
      "type": "integer",
      "description": "Number of resolution-pyramid levels. If set to `5`, there will be the full-resolution level and 4 levels of downsampled images."
    },
    "coarsening_xy": {
      "title": "Coarsening Xy",
      "default": 2,
      "type": "integer",
      "description": "Linear coarsening factor between subsequent levels. If set to `2`, level 1 is 2x downsampled, level 2 is 4x downsampled etc."
    },
    "image_extension": {
      "title": "Image Extension",
      "default": "tif",
      "type": "string",
      "description": "Filename extension of images (e.g. `\"tif\"` or `\"png\"`)."
    },
    "metadata_table_files": {
      "title": "Metadata Table Files",
      "type": "object",
      "additionalProperties": {
        "type": "string"
      },
      "description": "If `None`, parse Yokogawa metadata from mrf/mlf files in the input_path folder; else, a dictionary of key-value pairs like `(acquisition, path)` with `acquisition` a string like the key of the `acquisitions` dict and `path` pointing to a csv file containing the parsed metadata table."
    },
    "overwrite": {
      "title": "Overwrite",
      "default": false,
      "type": "boolean",
      "description": "If `True`, overwrite the task output."
    }
  },
  "required": [
    "acquisitions"
  ],
  "definitions": {
    "Window": {
      "title": "Window",
      "description": "Custom class for Omero-channel window, based on OME-NGFF v0.4.",
      "type": "object",
      "properties": {
        "min": {
          "title": "Min",
          "type": "integer",
          "description": "Do not change. It will be set to `0` by default."
        },
        "max": {
          "title": "Max",
          "type": "integer",
          "description": "Do not change. It will be set according to bit-depth of the images by default (e.g. 65535 for 16 bit images)."
        },
        "start": {
          "title": "Start",
          "type": "integer",
          "description": "Lower-bound rescaling value for visualization."
        },
        "end": {
          "title": "End",
          "type": "integer",
          "description": "Upper-bound rescaling value for visualization."
        }
      },
      "required": [
        "start",
        "end"
      ]
    },
    "OmeroChannel": {
      "title": "OmeroChannel",
      "description": "Custom class for Omero channels, based on OME-NGFF v0.4.",
      "type": "object",
      "properties": {
        "wavelength_id": {
          "title": "Wavelength Id",
          "type": "string",
          "description": "Unique ID for the channel wavelength, e.g. `A01_C01`."
        },
        "index": {
          "title": "Index",
          "type": "integer",
          "description": "Do not change. For internal use only."
        },
        "label": {
          "title": "Label",
          "type": "string",
          "description": "Name of the channel."
        },
        "window": {
          "$ref": "#/definitions/Window",
          "title": "Window",
          "description": "Optional `Window` object to set default display settings for napari."
        },
        "color": {
          "title": "Color",
          "type": "string",
          "description": "Optional hex colormap to display the channel in napari (it must be of length 6, e.g. `00FFFF`)."
        },
        "active": {
          "title": "Active",
          "default": true,
          "type": "boolean",
          "description": "Should this channel be shown in the viewer?"
        },
        "coefficient": {
          "title": "Coefficient",
          "default": 1,
          "type": "integer",
          "description": "Do not change. Omero-channel attribute."
        },
        "inverted": {
          "title": "Inverted",
          "default": false,
          "type": "boolean",
          "description": "Do not change. Omero-channel attribute."
        }
      },
      "required": [
        "wavelength_id"
      ]
    },
    "MultiplexingAcquisition": {
      "title": "MultiplexingAcquisition",
      "description": "Input class for Multiplexing Cellvoyager converter",
      "type": "object",
      "properties": {
        "image_dir": {
          "title": "Image Dir",
          "type": "string",
          "description": "Path to the folder that contains the Cellvoyager image files for that acquisition and the MeasurementData & MeasurementDetail metadata files."
        },
        "allowed_channels": {
          "title": "Allowed Channels",
          "type": "array",
          "items": {
            "$ref": "#/definitions/OmeroChannel"
          },
          "description": "A list of `OmeroChannel` objects, where each channel must include the `wavelength_id` attribute and where the `wavelength_id` values must be unique across the list."
        }
      },
      "required": [
        "image_dir",
        "allowed_channels"
      ]
    }
  }
}
```

Parallel:
```
{
  "title": "CellvoyagerToOmeZarrCompute",
  "type": "object",
  "properties": {},
  "required": [],
  "definitions": {
    "InitArgsCellVoyager": {
      "title": "InitArgsCellVoyager",
      "description": "Arguments to be passed from cellvoyager converter init to compute",
      "type": "object",
      "properties": {
        "image_dir": {
          "title": "Image Dir",
          "type": "string"
        },
        "plate_prefix": {
          "title": "Plate Prefix",
          "type": "string"
        },
        "well_ID": {
          "title": "Well Id",
          "type": "string"
        },
        "image_extension": {
          "title": "Image Extension",
          "type": "string"
        },
        "image_glob_patterns": {
          "title": "Image Glob Patterns",
          "type": "array",
          "items": {
            "type": "string"
          }
        },
        "acquisition": {
          "title": "Acquisition",
          "type": "integer"
        }
      },
      "required": [
        "image_dir",
        "plate_prefix",
        "well_ID",
        "image_extension"
      ]
    }
  }
}
```

zonia3000:
## Checklist before merging

Closes #456
Closes #451
Closes #445
Closes #468
Closes #469
Closes #470

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -10,7 +10,7 @@ jobs:
   end_to_end_tests:
     name: 'Node ${{ matrix.node-version }}'
     runs-on: ubuntu-22.04
-    timeout-minutes: 10
+    timeout-minutes: 20
 
     strategy:
       matrix:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
     * set default first task when continuing a workflow (\#466);
     * displayed applied filters in workflow execution modal (\#466);
     * implemented import and export of datasets (\#467);
+    * handled selection of default dataset on workflow page (\#467);
 
 # 0.10.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
     * implemented continue/restart workflow (\#465);
     * set default first task when continuing a workflow (\#466);
     * displayed applied filters in workflow execution modal (\#466);
+    * implemented import and export of datasets (\#467);
 
 # 0.10.2
 
--- a/__tests__/v2/workflow_utilities.test.js
+++ b/__tests__/v2/workflow_utilities.test.js
@@ -1,0 +1,75 @@
+import { describe, it, expect } from 'vitest';
+import {
+	deleteDatasetSelectionsForProject,
+	getDefaultWorkflowDataset,
+	getSelectedWorkflowDataset,
+	saveSelectedDataset
+} from '$lib/common/workflow_utilities.js';
+
+describe('Workflow Utilities', () => {
+	it('Default dataset is taken from most recent job', () => {
+		const id = getDefaultWorkflowDataset(
+			[],
+			[
+				{ start_timestamp: '2024-04-24T10:18:37.868655+00:00', dataset_id: 1 },
+				{ start_timestamp: '2024-04-24T15:18:37.868655+00:00', dataset_id: 2 }
+			]
+		);
+		expect(id).toEqual(2);
+	});
+
+	it('Default dataset is taken from most recent dataset', () => {
+		const id = getDefaultWorkflowDataset(
+			[
+				{ id: 1, timestamp_created: '2024-04-24T10:18:37.868655+00:00' },
+				{ id: 2, timestamp_created: '2024-04-24T15:18:37.868655+00:00' }
+			],
+			[]
+		);
+		expect(id).toEqual(2);
+	});
+
+	it('Default dataset is undefined if there are no jobs and no datasets', () => {
+		const id = getDefaultWorkflowDataset([], []);
+		expect(id).toEqual(undefined);
+	});
+
+	it('Store and retrieve selected dataset from localStorage', () => {
+		const workflow1 = { id: 1, project_id: 1 };
+		const workflow2 = { id: 2, project_id: 1 };
+		const datasets = [{ id: 1 }, { id: 2 }, { id: 3 }];
+		let datasetId = getSelectedWorkflowDataset(workflow1, datasets, 3);
+		// return default dataset if nothing is stored in localStorage
+		expect(datasetId).toEqual(3);
+
+		saveSelectedDataset(workflow1, 2);
+		datasetId = getSelectedWorkflowDataset(workflow1, datasets, 3);
+		// return dataset saved in localStorage
+		expect(datasetId).toEqual(2);
+
+		saveSelectedDataset(workflow1, 4); // dataset doesn't exist
+		datasetId = getSelectedWorkflowDataset(workflow1, datasets, 3);
+		// return default dataset if the dataset stored in localstorage doesn't exist
+		expect(datasetId).toEqual(3);
+
+		saveSelectedDataset(workflow1, 1); // replace stored selection
+		datasetId = getSelectedWorkflowDataset(workflow1, datasets, 3);
+		// return dataset saved in localStorage
+		expect(datasetId).toEqual(1);
+
+		saveSelectedDataset(workflow2, 2); // add new selection
+		datasetId = getSelectedWorkflowDataset(workflow2, datasets, 3);
+		// return dataset saved in localStorage
+		expect(datasetId).toEqual(2);
+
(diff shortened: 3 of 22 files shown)


tcompa:
I tested a few behaviors, and this works as expected.

One minor suggestion: let's sort datasets in the "select dataset" dropdown menu.

PR #481, opened 2024-05-07
Improved FormBuilder component and dataset history modal

### LINKED ISSUE #197: Delete a WorkflowTask argument (when not using JSON Schemas)
NOTE: this issue applies to the "legacy" free-arguments component, not to the one based on JSON Schemas.

See https://github.com/fractal-analytics-platform/fractal-server/issues/629#issuecomment-1600430008:
> As of #758, fractal-server now allows deleting an argument via the PATCH-WorkflowTask endpoint.
>
> If the current `WorkflowTask.args` is `{"a": 1, "b": 2}`, then sending a PATCH request with `args={"a": 3}` will lead to a new `WorkflowTask.args = {"a":3}` (without `b`).
>
> This behavior is available for all arguments that do not have a default value in the task args_schema (or for all arguments of tasks without args_schema). If a task has an args_schema, and `b` has a default, then with the same procedure as above we'll end up with `args = {"a": 3, "b": DEFAULT_OF_B}`.
>
> It is not possible to fully remove an argument if it has a default value in the schema, and we are not planning to introduce this feature for the moment.

This means that as of fractal-server 1.3.0a9 we can implement a feature to remove an argument.

### LINKED ISSUE #317: Use collapsible sections in Dataset.history
As of v0.6.1, this looks like
image

Maybe we can turn each history item in a collapsed-by-default section? That way, the user would just expand relevant items to look at them in detail.

zonia3000:
## Checklist before merging

Closes #317
Closes #197


tcompa:
I tried a fairly complex set of arguments
image
and the exported JSON seems correct
```json
{
  "args_parallel": {
    "asd": {
      "32": true,
      "asd": "asd",
      "array": [
        true,
        false,
        {
          "key": "value"
        }
      ]
    }
  },
  "args_non_parallel": {
    "key1": "val",
    "key2": 1,
    "key3": true
  }
}
```

I also tried to change page in a few different ways, and I was prompted the save/cancel/discard dialog.

All seems to work as expected.

---

I had a look at at the new collapsible-items history, and it looks good to me.

tcompa:
Minor glitch: I think that "Argument default value" should be "Argument value" (with no "default")

image

tcompa:
Minor glitch: we call the same thing "Argument", "Parameter" and "Property".

image

Should we drop at least one of those three names? I suggest we keep "property" (because it's also used in the nested objects), but drop one of "Arguments" and "Parameters".

tcompa (approved):

PR #485, opened 2024-05-14
Included v1/v2 switcher in admin-area sub-pages

### LINKED ISSUE #478: Include v1/v2 switcher in all admin-area pages
As an admin, it's useful to go back and forth between v1 and v2 admin-area jobs pages (e.g. to check whether something is running). Right now, it requires several steps (going back to the home page and then to the legacy section and then to the admin area and then to the jobs page). I suggest we include the switcher in all admin-area pages.

### LINKED ISSUE #486: Remove "users" page from v1 admin-area

zonia3000:
## Checklist before merging

Closes #478
Closes #486

The switcher is not visible in the admin pages that exists only in v2: tasks and tasks-compatibility.


tcompa:
> It works as expected, thanks!
> 
> Perhaps we could also include #486, and avoid duplicating the V1 admin-area page for users (since the two pages are strictly identical, I think).

This means that the switcher would only be on the "jobs" page, which was anyway the most relevant use case for #478.

PR #490, opened 2024-05-17
Improved JSON Schema Sandbox page

### LINKED ISSUE #418: Improve JSON-schema sandbox
This is very low priority, but it will be useful the next time we need to play with argument JSON schemas.

1. Could we access the sandbox page without authentication? We can decide to hide it unless a specific env variable is set (are we already doing it?)
2. Right now I think there is a glitch in what is considered as the "data" field value. As a first fix: can we always set the data field to `{}` or to empty, at page refresh? In the current version, a page refresh seems to reset the schema prompt but not the data one.
3. Let's increase the height of the JSON-Schema prompt, because it obviously never fits in a single line. Maybe some 10-20 lines, with a scrollbar? Or, even better, let's place the schema and data prompts on the left an the form UI on the right.
4. (a bit more testing will probably raise more of these minor issues)

The overall idea is that most activities in the https://github.com/fractal-analytics-platform/fractal-web/issues?q=is%3Aissue+is%3Aopen+label%3A%22JSON+Schemas%22 area should be easy to test within this sandbox.

zonia3000:
## Checklist before merging

Closes #418


tcompa:
This works as expected, thanks!
A couple of possible improvements, in case they are not too complex:
1. Can we include a "load example" button? The longest part of testing, for me, was to find a JSON Schema to play with. A valid example could be any JSON schema from the fractal-tasks-core ones, or even a very simple one generated on purpose (with a couple of scalar property, an array property, an object property, and some properties marked as required).
2. When validating, could we also display the actual data that are being validated (or otherwise we could add a "show current data" button)? This would be useful to understand e.g. how we are handling `null` values, or empty strings.

tcompa (approved):

zonia3000:
image

I've added a "Load example" button and a "Load current data" button that retrieves the values set using the form and load them inside the JSON textarea.

PR #504, opened 2024-06-03
Added maintenance banner when fractal-server is down

### LINKED ISSUE #491: Display maintenance page/banner when backend is down

### LINKED ISSUE #501: Question about "immutable/chunks" logs
The logs show records like
```
2024-05-30 14:29:00,630 - (hooks) - INFO - [GET] - /_app/immutable/nodes/33.d1a587dd.js
2024-05-30 14:29:00,641 - (hooks) - INFO - [GET] - /_app/immutable/chunks/taskStores.e8636c18.js
2024-05-30 14:29:00,645 - (hooks) - INFO - [GET] - /v2/tasks
2024-05-30 14:29:00,646 - (hooks) - INFO - [GET] - /_app/immutable/chunks/TasksTable.194b98d0.js
```

@zonia3000 could you explain briefly what they represent?

I guess they are expected, since those are INFO logs - is there any situation where this kind of logs require some further check?

zonia3000:
## Checklist before merging

Closes #491
Closes #501

This is the banner appearing when the server is down:

image

Some notes about the changes.

I moved the logic that was inside `+layout.server.js` to `hooks.server.js`. This is because we want to check if the server is down before performing further calls to the API. The banner is displayed in all public pages. If one tries to access a private page and the server is down the user is redirected to the home page, where the banner can be displayed. This is needed to avoid performing the calls in any `+page.server.js`, that would result in a 500 error when the server is down.

I've also retrieved the current user only in `hooks.server.js` (we were retrieving it both in `hooks.server.js` and `+layout.server.js`). The object `locals` is used to pass values (user info and server info) from `hooks.server.js` to `+layout.server.js` (see [Svelte doc](https://kit.svelte.dev/docs/types#app-locals) or [this more comprehensive guide](https://khromov.se/the-comprehensive-guide-to-locals-in-sveltekit/)).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,6 +7,7 @@ node_modules
 .env.*
 !.env.example
 *.d.ts
+/venv
 
 # Ignore files for PNPM, NPM and YARN
 pnpm-lock.yaml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
+# Unreleased
+
+* Added maintenance banner when fractal-server is down (\#504).
+
 # 1.1.0
 
 > WARNING: with this release all the environment variables will be read from
--- a/src/hooks.server.js
+++ b/src/hooks.server.js
@@ -23,7 +23,7 @@ export async function handle({ event, resolve }) {
 		// reach this point if a not existing file is requested. That could happen after an update
 		// if the browser is loading a cached page that references to outdated contents.
 		// In that case we can just skip the whole function.
-		await resolve(event);
+		return await resolve(event);
 	}
 
 	if (event.url.pathname.startsWith('/api')) {
--- a/src/hooks.server.js
+++ b/src/hooks.server.js
@@ -17,19 +17,24 @@ if (env.FRACTAL_SERVER_HOST.endsWith('/')) {
 }
 
 export async function handle({ event, resolve }) {
-	if (event.url.pathname.startsWith('/_app')) {
-		// The _app folder contains static files that are the result of npm build.
-		// The hooks handle function is not called when loading valid static files, however we can
-		// reach this point if a not existing file is requested. That could happen after an update
-		// if the browser is loading a cached page that references to outdated contents.
-		// In that case we can just skip the whole function.
-		return await resolve(event);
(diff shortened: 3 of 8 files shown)


tcompa:
The maintenance page looks as expected - thanks!

---

Concerning #501, I tried to stress-test it by removing the whole content of `build/client/_app/immutable/chunks/` (which looked like a fake way to invalidating a cache - sort of). Then I continued interacting with the app, and got this kind of error
```
2024-06-05 05:55:07,107 - (hooks) - INFO - [GET] - /auth/login
2024-06-05 05:55:09,416 - (hooks) - INFO - [GET] - /v2/jobs
2024-06-05 05:55:09,451 - (hooks) - INFO - [GET] - /v2/jobs
2024-06-05 05:55:33,301 - (page layout) - ERROR - pageInfo is missing, it should have been loaded by hooks.server.js
NotFound [Error]: Not found: /_app/immutable/nodes/19.ad3f3466.js
    at resolve (/somewhere/Fractal/fractal-web/node_modules/@sveltejs/kit/src/runtime/server/respond.js:483:13)
    at resolve (/somewhere/Fractal/fractal-web/node_modules/@sveltejs/kit/src/runtime/server/respond.js:277:5)
    at Object.handle (/somewhere/Fractal/fractal-web/src/hooks.server.js:26:9)
    at Module.respond (/somewhere/Fractal/fractal-web/node_modules/@sveltejs/kit/src/runtime/server/respond.js:274:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 404
}
2024-06-05 05:55:33,319 - (hooks) - INFO - [GET] - /v2/admin
2024-06-05 05:55:33,431 - (hooks) - INFO - [GET] - /_app/immutable/nodes/19.ad3f3466.js
```

Is it correct that the last INFO line still mentions "_app/immutable"?

zonia3000:
@tcompa I've found a better solution. The event object inside the hooks has a property route, which has a null id if there is no Svelte route associated with that path. In that case we can just skip the whole execution using:

```javascript
throw error(404, 'Route not found');
```

The result is a standard Svelte 404 page:

image

Log is also clearer, because handling it explicitly we avoid the Svelte stack trace about the missing file.

Notice that this doesn't affects other 404 errors, that we are handling explicitly, like the one in projects:

image

PR #509, opened 2024-06-07
JSON Schema form improvements

### LINKED ISSUE #503: `enum` support within `definitions`
I have two JSON Schemas, both making use of `definitions`

The first one
```json
{
  "title": "TaskFunction1",
  "type": "object",
  "properties": {
    "arg1": {
      "$ref": "#/definitions/InitArgsRegistration",
      "title": "Arg1",
      "description": "Description of arg1."
    }
  },
  "required": [
    "arg1"
  ],
  "definitions": {
    "InitArgsRegistration": {
      "title": "InitArgsRegistration",
      "description": "Registration init args.",
      "type": "object",
      "properties": {
        "reference_zarr_url": {
          "title": "Reference Zarr Url",
          "type": "string"
        }
      },
      "required": [
        "reference_zarr_url"
      ]
    }
  }
}
```
renders correctly in the sandbox
image

---

The second one
```json
{
  "title": "TaskFunction2",
  "type": "object",
  "properties": {
    "arg2": {
      "$ref": "#/definitions/Color",
      "title": "Arg2",
      "description": "Description of arg2."
    }
  },
  "required": [
    "arg2"
  ],
  "definitions": {
    "Color": {
      "title": "Color",
      "description": "An enumeration.",
      "enum": [
        1,
        2
      ]
    }
  }
}
```
does not render:
image
and dev-tools log do not show anything special
image

---

As a double check, the second schema does render on https://rjsf-team.github.io/react-jsonschema-form/:

image

---

What is that makes our current form creation fail?
Is it because in the second one the definition of `Color` does not have an additional nested `properties` property?

### LINKED ISSUE #505: Showing tuple parameters in workflow task page
I'm not 100% sure whether this is a manifest creation or Fractal web display issue. When a Fractal task has an input that's a tuple with multiple values, it's not displayed in the Workflow task.
The manifest looks reasonable to me, so I assume that's an issue with how we than handle the schema in Fractal web.

Example task parameters:
```
@validate_arguments
def rdcnet_segmentation_2d(
    *,
    # Fractal arguments
    input_paths: Sequence[str],
    output_path: str,
    component: str,
    metadata: Dict[str, Any],
    # Task-specific arguments
    level: int,
    channel: ChannelInputModel,
    model_path: str,
    input_ROI_table: str = "FOV_ROI_table",
    output_ROI_table: Optional[str] = None,
    output_label_name: Optional[str] = None,
    use_masks: bool = False,
    relabeling: bool = True,
    # RDCNet-related arguments
    core_count: int = 4,
    # 2D RDCNet-related arguments
    patch_size: tuple[int, int, int] = [1300, 1300, 1],         <= the relevant parameter
    overwrite: bool = True,
) -> Dict[str, Any]:

```

The relevant entry in the manifest:
``` 
"patch_size": {
      "title": "Patch Size",
      "default": [
        1300,
        1300,
        1
      ],
      "type": "array",
      "minItems": 3,
      "maxItems": 3,
      "items": [
        {
          "type": "integer"
        },
        {
          "type": "integer"
        },
        {
          "type": "integer"
        }
      ],
      "description": "Size of the patch, excluding border, see RDCNet documentation."
    },
```

How it gets displayed in Fractal web:

Screenshot 2024-06-04 at 21 28 01

The default values aren't shown and one can not enter actual values, just add or remove items from the list.

------

@FranziskaMoos-FMI : I looked back at how I solved this in my original version of the task: I used list[int]. It's not optimal, as a user could mistakenly use a wrong number of elements, but it is displayed. See https://github.com/fmi-basel/isit/blob/044c66490de4f727087aac3c715226ab3f581ee1/isit/fractal/rdcnet_segmentation.py#L108

### LINKED ISSUE #506: Add info about JSON Schema sandbox in docs
I think this is currently missing.
Also, we should specify that when this page is accessible (e.g. the difference between `npm run dev` and `npm run build; npm run preview`).

### LINKED ISSUE #507: Display Optional Arguments that are `None` in grey
Lorenzo and I just started using the sandbox for Json schemas. @zonia3000 super useful to test things!! 🎉

One idea that came up: Can we highlight better when Optional values are unset, e.g. when they are still None? The idea we had: If an Optional field is None (e.g. empty string), it is greyed out. If the user fills it in, it becomes regular black.

One example: The Cellpose task has an Optional `Output Roi Table` parameter. If it's unset, it could be greyed out, signaling to the user that this parameter isn't currently being used.

Screenshot 2024-06-05 at 15 31 40

Can we experiment with that?

### LINKED ISSUE #510: Add new sandbox page for the version-update feature
Since there the JSON-schema building may change a bit in the future (see https://github.com/fractal-analytics-platform/fractal-tasks-core/labels/JSON%20Schemas), it would be useful to be able to test the version-update page in a sandbox, as we can already do for the form generation.

A possible structure could be similar to the current sandbox:
1. A text prompt for the old schema, with the generated UI on its right and the current values shown somewhere.
2. A text prompt for the new schema.
3. A button like "try version update" that triggers the same steps as in the current version-update user flow (or very similar ones).

zonia3000:
## Checklist before merging

Closes #507
Closes #506
Closes #505
Closes #503
Closes #510


tcompa:
As we just discussed:
- When running in development mode, let's include some browsing links for the sandboxes (one to get there from the home page, and then links to move between the two sandboxes)
- Within the version sandbox, let's also validate that the new JSON Schema is valid.

tcompa:
There is a small glitch when using nested tuples as in 
```json
{
  "title": "TaskFunction",
  "type": "object",
  "properties": {
    "arg_A": {
      "title": "Arg A",
      "type": "array",
      "minItems": 1,
      "maxItems": 1,
      "items": [
        {
          "type": "array",
          "minItems": 1,
          "maxItems": 1,
          "items": [
            {
              "type": "array",
              "minItems": 1,
              "maxItems": 1,
              "items": [
                {
                  "type": "string"
                }
              ]
            }
          ]
        }
      ],
      "description": "Description of arg_A."
    }
  },
  "required": [
    "arg_A"
  ],
  "additionalProperties": false
}
```

where the non-top-level tuples still have the "Add argument to list" button:
image

tcompa (approved):

tcompa:
A minor addition that could be helpful is to include a "load example" button in the sandbox/versions page (similar to the one in sandbox/jsonschema). It can be a very simple one (e.g. an property-name change, or a new required property).

zonia3000:
> There is a small glitch when using nested tuples as in

Now fixed:

image

zonia3000:
I've added a Sanbox button in home page (visible only in development mode):

image

It brings to a page showing the list of sandbox pages and adds a "Sandbox pages" link in the top menu, so that one can navigate back to the list easily:

image

---

This is the content loaded by the "Load example" in version sandbox:

image

---

JSON Schema is now validated also in sandbox:

image

tcompa (approved):

PR #518, opened 2024-06-28
Refactor of JSchema component; moved Sandbox pages to docs

zonia3000:
## Checklist before merging


tcompa:
I observed something which looks like a small bug.

I am working with a task with the following parallel JSON Schema:
```json
{
  "title": "IlluminationCorrection",
  "type": "object",
  "properties": {
    "zarr_url": {
      "title": "Zarr Url",
      "type": "string",
      "description": "Path or url to the individual OME-Zarr image to be processed. (standard argument for Fractal tasks, managed by Fractal server)."
    },
    "illumination_profiles_folder": {
      "title": "Illumination Profiles Folder",
      "type": "string",
      "description": "Path of folder of illumination profiles."
    },
    "illumination_profiles": {
      "title": "Illumination Profiles",
      "type": "object",
      "additionalProperties": {
        "type": "string"
      },
      "description": "Dictionary where keys match the `wavelength_id` attributes of existing channels (e.g. `A01_C01` ) and values are the filenames of the corresponding illumination profiles."
    },
    "background": {
      "title": "Background",
      "default": 0,
      "type": "integer",
      "description": "Background value that is subtracted from the image before the illumination correction is applied. Set it to `0` if you don't want any background subtraction."
    },
    "input_ROI_table": {
      "title": "Input Roi Table",
      "default": "FOV_ROI_table",
      "type": "string",
      "description": "Name of the ROI table that contains the information about the location of the individual field of views (FOVs) to which the illumination correction shall be applied. Defaults to \"FOV_ROI_table\", the default name Fractal converters give the ROI tables that list all FOVs separately. If you generated your OME-Zarr with a different converter and used Import OME-Zarr to generate the ROI tables, `image_ROI_table` is the right choice if you only have 1 FOV per Zarr image and `grid_ROI_table` if you have multiple FOVs per Zarr image and set the right grid options during import."
    },
    "overwrite_input": {
      "title": "Overwrite Input",
      "default": true,
      "type": "boolean",
      "description": "If `True`, the results of this task will overwrite the input image data. If false, a new image is generated and the illumination corrected data is saved there."
    },
    "suffix": {
      "title": "Suffix",
      "default": "_illum_corr",
      "type": "string",
      "description": "What suffix to append to the illumination corrected images. Only relevant if `overwrite_input=False`."
    }
  },
  "required": [
    "zarr_url",
    "illumination_profiles_folder",
    "illumination_profiles"
  ],
  "additionalProperties": false
}
```

I open the form and remove one entry from the "illumination profiles" property.
I click "save", and I get a weird unrelated error:
image

---

I cannot reproduce within the sandbox. I'll try to debug further.

tcompa:
If I export the arguments, I see that `overwrite_input` is a string rather than a boolean:
```json
{
  "args_parallel": {
    "background": 110,
    "input_ROI_table": "FOV_ROI_table",
    "overwrite_input": "True",
    "suffix": "_illum_corr",
    "illumination_profiles_folder": "/home/fractal_share/fractal-demos-v2/examples/02_cardio_small/../illum_corr_images/",
    "illumination_profiles": {
      "A01_C01": "20220621_UZH_manual_illumcorr_40x_A01_C01.png",
      "A01_C02": "20220621_UZH_manual_illumcorr_40x_A01_C02.png",
      "A02_C03": "20220621_UZH_manual_illumcorr_40x_A02_C03.png"
    }
  },
  "args_non_parallel": null
}
```

tcompa:
This is possibly due to a different issue (https://github.com/fractal-analytics-platform/fractal-demos/issues/59). I'll review it there.

tcompa:
I confirm it was a fractal-demos issue (see https://github.com/fractal-analytics-platform/fractal-demos/pull/79/files).

tcompa:
@jluethi: there has been yet another update with respect to our latest discussion. We finally moved away from Web Components, since there exist a much simpler way to include static html pages in the docs.

A recap of the new repo/package organization as far as I understand it (see also new docs introduced in 62dddb6126fa7973be1283525a3af7ed2d95785f):
* We now have three Svelte projects in the repo: `fractal-web`, `jschema` and `sandbox`.
* `fractal-web` development and releases do not change (apart from some minor caveat for developers e.g. about relative imports from `jschema`)
* All the (refactored) logic of the form-generator is now part of `jschema`.
* `jschema` is not a released package (that is, we are not introducing two versioned projects: only fractal-web has versions)
* `sandbox` is an independent project, with no connection with `fractal-web` (apart from the fact that they both import from `jschema`). Also `sandbox` has no specified version (it is rather alwasy bound to the `fractal-web` one).
* `sandbox` can be run locally with usual tools (e.g. `npm run dev` from the `sandbox` folder), which is the typical use for developmente
* `sandbox` is built as part of the docs GitHub action and its static files are included in the online docs
* `sandbox` is never exposed within `fractal-web` in a production instance
* Because of this new approach (not using Web Components), the issues with bootstrap JS are not present any more.

tcompa:
I've played with this new version, and found nothing unexpected. I suggest we merge into `main` (so that we can also review the new online docs).

**NOTE** that the only expected behavior change is the bugfix for #520 which -as a side effect- removes the smooth animation of opening/closing an accordion. Once again: this is *not* any more a consequence of issues with bootstrap js, but rather a consequence of the bugfix.
I think this change of behavior is OK, but let's open a separate issue in case we prefer to keep the animation smooth.

tcompa (approved):

jluethi:
This is awesome, thanks both! And thanks for the informative summary!

Re animation change: I looked at it now. Totally fine by :)

PR #521, opened 2024-07-02
Added Task Manifest Sandbox page, validated text in numeric inputs

### LINKED ISSUE #513: Parse whole task-package manifest, and show argument forms for tasks
This is a placeholder issue, which will become relevant during/after #512.

Briefly: we could let the user upload a whole manifest (anything upload the file, paste the JSON text, point to an URL), parse it to extract the N tasks and their parallel/non-parallel schemas, and show e.g. a dropdown menu with all tasks. By selecting one, the user would see the corresponding form and values.

High-level: this would be quite useful for a user who built their package and wants to review the whole manifest (are tasks all there? do they look OK in the form? and so on..). The parsing of the manifest is quite simple, and based on this JSON Schema https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/json_schemas/manifest_v2.json

### LINKED ISSUE #516: Do we need "Clear" for elements of "tuple" properties?
I don't have a very strong opinion here, but the issue is to collect feedback about the "clear" button when we have a fixed-length array (as raised by @lorenzocerrone).

If an array must have 3 elements, and we "clear" one of them, the current data has only two elements and validation will obviously fail (as it does, correctly). What is the use case for the "clear" button then?

[maybe it's simply a consequence of developing the "tuple" UI based on the original array UI, which required dynamic addition/removal of elements]

Feedback or examples are welcome

### LINKED ISSUE #519: Invalid value in form field is silently discarded upon save
When an optional field is set to something invalid (e.g. characters instead of digits), we show a small error message as expected:

image

However one can still click "save changes", and the validation and backend calls will take place with the corresponding field unset.

Is this the expected behavior, or should we raise an error when clicking "save changes" if there are some invalid fields?

zonia3000:
## Checklist before merging

Closes #513
Closes #519
Closes #516


tcompa:
V1 manifests are not currently supported:
image
and I think this is the correct behavior.

Thus we should drop the "Legacy: Changes the set of ignored properties (v1 or v2)" switcher

tcompa:
Something is not working re: the non-parallel/parallel switch. The first task of fractal-tasks-core has both schemas ([non-parallel](https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/fractal_tasks_core/__FRACTAL_MANIFEST__.json#L16) and [parallel](https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/fractal_tasks_core/__FRACTAL_MANIFEST__.json#L181)), but the choice of "parallel" is disabled:
image

tcompa:
As @zonia3000 correctly pointed out, form generation for compound tasks works fine - see e.g.
image


The reason of not seeing the parallel task form in the previous comment is that the corresponding schema is empty (once we strip the default properties which are handled by fractal-server).

tcompa:
I also reviewed #521, and it works as expected:
image

jluethi:
Awesome! 🤩

Let's keep it to V2 only for sure :)

tcompa (approved):

PR #522, opened 2024-07-08
Fixed ajv formats error on sandbox

zonia3000:
## Checklist before merging
- [X] I added an appropriate entry to `CHANGELOG.md`

PR #535, opened 2024-07-24
Supported prefixItems (JSON Schema Draft 2020-12 - pydantic_v2)

### LINKED ISSUE #534: Adapt to JSON Schema 2020-12 changes in `items`
There is at least one more issue related to #531, which will require a bit more work (i.e. handling/testing `pydantic_v1` and `pydantic_v2` schemas in different ways).

---

Pydantic V2 generates JSON Schemas that comply with draft 2020-12, which means that they follow the new style for items: https://json-schema.org/draft/2020-12/release-notes#changes-to-items-and-additionalitems. Briefly
> The `items` and `additionalItems` keywords have been replaced with `prefixItems` and `items` where `prefixItems` has the same functionality as the array-of-schemas for of the old `items` and the new `items` keyword has the same functionality as the old `additionalItems` keyword.

---

Starting from a task Python function with an argument `arg_A: tuple[int, int] = (1, 1)`, we end up with this kind of JSON Schema:
```json
{
  "properties": {
    "arg_A": {
      "default": [
        1,
        1
      ],
      "maxItems": 2,
      "minItems": 2,
      "prefixItems": [
        {
          "type": "integer"
        },
        {
          "type": "integer"
        }
      ],
      "title": "Arg A",
      "type": "array",
      "description": "Description of arg_A."
    }
  },
  "type": "object",
  "title": "TaskFunction"
}
```
which currently cannot be loaded in fractal-web, due to `error compiling schema Error: strict mode: unknown keyword: "prefixItems"`.

---

Since we need to tackle this question on the form-generation side, let's also check whether we should configure ajv differently for the two `pydantic_v1/v2` cases, that is by specifying the JSON Schema draft explicitly.

### LINKED ISSUE #536: Tasks should not loose `input filters` upon updating them
Reported by @nrepina :

> I noticed when I update task versions on fractal web, the arguments and metadata update nicely, but the ‘input filters’ are cleared, so I need to re-specify them again for all tasks that are updated (e.g. is_3D=True). It’s a little annoying, so if the default can be to keep the type filters from the previous version, what would be great

That's a bug in task updating we should fix. The user expectation should be that those filters are not affected by the update.

### LINKED ISSUE #540: Extend default handling from undefined-only to null/empty argument objects
Given https://github.com/fractal-analytics-platform/fractal-web/issues/538, we have case A (arguments object is "empty" and fractal-web fills the defaults) and case B (arguments objects has values, and fractal-web ignore defaults).

Right now, case A actually requires arguments to be "undefined". We should extend it to cases where the object exists but is either `null` or `{}`.

This will also make the sandbox behavior (where `undefined` is sometimes found) more similar to the webclient (where the backend always sends something for the arguments property).

zonia3000:
## Checklist before merging

Closes #534
Closes #536
Closes #540


zonia3000:
According to [Ajv documentation](https://ajv.js.org/json-schema.html#draft-2020-12) to use the draft-2020-12 it is necessary to load a different Ajv instance.

image

Regarding Ajv I've also discovered that the library behaves differently in dev or production mode when using `import Ajv from 'ajv'` (i.e. strict mode doesn't work in dev), while the library has the same behavior both in dev or production mode when imported using `import Ajv from 'ajv/dist/ajv';`. I'm not sure what is causing that, but changing the import seemed to produce a consistent behavior. Maybe it is the same cause of #522, where we had a different behavior in dev/production too.

Supporting the 2020-12 draft requires to pass the schema version (`pydantic_v1` or `pydantic_v2`) to the `JSchema` component and its children. This also has an impact on sandbox pages, where I've added some radio buttons to select the schema versions:

image

image

The current PR assumes that Pydantic v2 generates `prefixItems` for tuples and `items` for arrays. We are not currently supporting a situation where an array has both `prefixItems` and `items`. I suppose that case might be possible when defining only `minItems` (so leaving max unlimited) with a specific format for the first n items and a different format for the remaining items. It doesn't seem very common.

tcompa:
The update should include also the create-single-task scenario:

image

tcompa:
In general, it would be obviously nice for the user if they don't have to explicitly know about pydantic v1/v2 (or JSON Schema draft version).

At least in the single-task creation, this can be achieved e.g. as follows:
1. Try to load as a 2020-12 draft schema
2. If it fails, try with an older draft
3. If it fails, raise an error

I don't know if the same pattern can be used in every place where we have this logic - but I'd suggest we do something similar when it's possible.

zonia3000:
We already have the `Args schema version` field, that appears when the user uploads an args schema. It's just that it appears after the check so it doesn't work for the v2. It is a bug that I can fix now.

image

The component can try to autodetect the version based on the 3 points you wrote and then fill the fields accordingly.
The user should be able to modify it because a v1 schema can be a valid v2 schema but with a different meaning.

tcompa:
I've tested single-task creation as of https://github.com/fractal-analytics-platform/fractal-web/pull/535/commits/414924ad9db7e7e1dd42273bd1b8da20ce3c4fc4, and it does work as expected.

I am using the following schema
```json
{
  "properties": {
    "arg_A": {
      "maxItems": 2,
      "minItems": 2,
      "prefixItems": [
        {
          "type": "integer"
        },
        {
          "type": "integer"
        }
      ],
      "title": "Arg A",
      "type": "array",
      "description": "Description of arg_A."
    },
    "arg_B": {
      "items": {
        "type": "integer"
      },
      "title": "Arg B",
      "type": "array",
      "description": "Description of arg_B."
    }
  },
  "required": [
    "arg_A",
    "arg_B"
  ],
  "type": "object",
  "title": "TaskFunction"
}
```
and adding it to a new task.

At first, I use the default proposed version `pydantic_v2`. All looks fine 
image

---

In a second test, I tried to modify the default proposed version into `pydantic_v1`, but the POST API call still has `args_schema_version="pydantic_v2"`. The reason I'd need the possibility to customize this field is that I want to see what UI will come out of a version mismatch.

tcompa:
Last minor comment (as already discussed): since the only supported versions are pydantic_v2 or v1, let's make this choice a dropdown menu with a preselected default option.

tcompa:
I confirm that https://github.com/fractal-analytics-platform/fractal-web/pull/535/commits/b7857ddda65d74d04ed496bc7568f88b41daa41c fixes the aforementioned glitch. I can select `pydantic_v1`, and then I get an error upon clicking "Create":

image

PR #564, opened 2024-09-25
User settings

### LINKED ISSUE #563: Align with fractal-server 2.6.0 and 2.6.1 (re: user settings)
In fractal-server 2.6, we will extract several attributes from the user table into a dedicated user-settings table, and we will also start including more attributes (which were currently global variables).

The main breaking changes in the API are in the `/auth/users/ routes.
1. Three properties (`slurm_user`, `slurm_accounts`, `cache_dir`) are not part of request/response bodies for GET/PATCH requests to `/auth/users/{id}/` and `/auth/current-user/` any more.
2. New GET/PATCH endpoints are available at `/auth/users/{id}/settings/` (for the admin) or `/auth/current-users/settings/` (for the user).
3. It's still TBD (but likely) whether some endpoints (e.g. `GET /auth/users/` or `GET /auth/users/{id}/`) will automatically include a property `user.settings`, or whether this will have to go through a second API call.
4. The response of GET/PATCH `/auth/current-user/` will not include a `settings` property, but it will be required to make a second API call to `/auth/current-user/settings/`
5. Overall, the scope of `PATCH /auth/current-user/` will decrease, or it may even be removed - TBD.

Most internal changes concern the SLURM-related job execution and task collection, which are likely not tested in fractal-web CI.

---

In the webclient, we need to make changes in two parts:

### Admin area

In the page where the admin can edit a single user, they should see two different forms: one to edit the user properties, and one to edit the user_settings one. Each form has a different "save" button.

### User area

We should split the "profile" page into two pages: profile page (which is essentially read-only) and settings page (where the user can modify some values).

There is a decision to be taken about which settings to expose in the PATCH-related form. The database user-setting table has several columns, but only a subset of them are relevant. Which subset is relevant depends on a backend variable (`FRACTAL_RUNNER_BACKEND`). If we duplicate this env variable in fractal-web, then we can show a different form based on its value. Other ideas that do not rely on duplicating this variable are welcome.

zonia3000:
## Checklist before merging

Closes #563


tcompa:
Could we add here an info message that says something like "User settings can be modified after creating the user"?

image

zonia3000:
image


Some screenshots for the SSH Slurm configuration:

image

image

PR #572, opened 2024-10-02
Handled special case of Zarr URL containing hash

zonia3000:
## Checklist before merging


tcompa:
The current status is that we are not pursuing the corresponding fractal-vizarr-viewer feature (ref https://github.com/fractal-analytics-platform/fractal-vizarr-viewer/issues/33), due to how less readable URLs would become.

Closing.

PR #640, opened 2024-11-13
Alignment with fractal-server 2.9.0

### LINKED ISSUE #638: Align with fractal-server 2.9.0
Fractal-server 2.9.0 simplifies the task-collection-related API, because it will be used for more flexible actions.

High-level: task collection will be one of three possible task-group "activities": collection, deactivation, reactivation. Each task group can be collected only once, but then it can be deactivated and reactivated any number of times.

Some relevant changes

1. `POST /api/v2/task/collect/pip/` remains the same, but it responds with a 202 (instead of 201) and with a [TaskGroupActivityV2Read object](https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/app/schemas/v2/task_group.py#L103-L113). This is similar to the pre-2.9.0 response, but it _removes_ the unstructured `data` property of the previous response in favor of explicit properties (e.g. `pkg_name`, `version`, `status). The `action` property in this case is always `"collect"`.
2. `GET /api/v2/task/collect/{state_id}/"` has been removed, in favor of a more flexible `GET /api/v2/task-group/activity/{task_group_activity_id}/`. Its use is the same: for a given ID, you can fetch the collection information.
3. A new `GET /api/v2/task-group/activity/` endpoint offers the (new) feature of being able to see all _your_ task-group activities. For the moment those could only have  `action="collect"`, but starting from 2.9.0 there will also be `deactivate` and `reactivate`. TBD where/how we want to expose this information.
4. Note that this new  endpoint has a bunch of query parameters: `pkg_name` (fuzzy), `status`, `action`, `timestamp_started_min`. I think it'd be good to replace the current logic ("show me the collections that are stored in the browser cache") with something like "show me the collections of the last week" (and possibly some more advanced controls somewhere - see below).
5. We still need to decide where to expose and track the deactivate/reactivate actions.
6. While `GET /api/v2/task-group/activity/` gives a user their own list of activities, we are now introducing the possibility for an admin to view and query all activities via `GET /admin/v2/task-group/activity/`. Since there will be more complex users' actions in this area, I think it'd be good to introduce a corresponding admin-tools page.

All parts concerning task-group activity and task collection are already in place in current fractal-server `main`, while the parts about deactivate/reactivate are in-progress (but they will also be part of 2.9.0).

### LINKED ISSUE #639: Update dataset-creation modal
1. If the user has a `project_dir` in their settings -> `zarr_dir` is optional, and the corresponding text field is not visible by default (e.g. it's hidden in an "Advanced options" block)
2. If the user has no `project_dir` in their settings -> `zarr_dir` is required and the corresponding text field is visible

zonia3000:
## Checklist before merging

Closes #638
Closes #639

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -77,6 +77,7 @@ jobs:
           FRACTAL_RUNNER_BACKEND: local_experimental
 
       - name: Archive test results
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: test-results-folder
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 # Unreleased
 
 * Alignment with fractal-server 2.9.0 (\#640):
-    * Updated task collections logic and created task group activities pages for standard users and administrators (\#640);
+    * Updated task collections logic and created task-group activities pages for standard users and administrators (\#640);
     * Removed active checkbox from task group edit modal (\#640);
     * Added Manage button to deactivate and reactivate task groups (\#640);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
+# Unreleased
+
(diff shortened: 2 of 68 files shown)


zonia3000:
Deactivate:

image

image

Reactivate:

image

PR #658, opened 2024-12-02
Added task filters, split tasks page in "tasks" and "tasks management"

### LINKED ISSUE #654: Display/use new task metadata
As of fractal-tasks-core v1.3.3 (now available) we have an example of a task package with [the new metadata](https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/app/models/v2/task.py#L43-L48) (`authors`, `modality`, `category` and `tags`).

We will now have to start introducing these attributes in several task views, and review how we can expose some filtering in some specific situations (notably the tasks page and the add-task-to-workflow modal).

For most of these attributes, relevant fractal-servers GET endpoints already have query parameters available (the only missing one is for `tags` - ref https://github.com/fractal-analytics-platform/fractal-server/issues/1863). Depending on how filtering will work, we can either fix this issue backend-side, or we can re-implement all filtering in fractal-web.

### LINKED ISSUE #656: Add a dedicated `alive` page/endpoint?
Something like
```console
$ curl http://localhost:5173/alive/
{"alive":true,"version":"1.11.0"}
```
would be useful in https://github.com/fractal-analytics-platform/fractal-healthcheck.

### LINKED ISSUE #657: Split standard-user tasks page into "tasks page" vs "tasks management page"
The management page would look like the current one.

The tasks page would look very much like the discussion in https://github.com/fractal-analytics-platform/fractal-web/issues/654.

### LINKED ISSUE #659: Include user-email in `v2/admin/task-groups/activities`
Right now this column is not there:

image

If it's possible with the current API-provided information, we should add a "User" column the same way as in `v2/admin/task-groups`.

### LINKED ISSUE #660: Sort users list in dropdown menu in `v2/admin/task-groups/activities`

### LINKED ISSUE #662: Latest task-collection activities goes last in the list
When we trigger a task collection, the corresponding activities appears as the last in the list:
image

While after a refresh its position changes:
image

Two questions:
* Could we sort the list based on the "start" timestamp? (more recent first)
* Could we always add a new entry to the list on the top (since it will always be the most recent one)?

zonia3000:
## Checklist before merging

Closes #654
Closes #656
Closes #657
Closes #660
Closes #659
Closes #662


zonia3000:
@tcompa Could we add some metadata to `fractal-tasks-mock`, so that I can easily test the filtering end-to-end?

zonia3000:
> Now available in fractal-server.

Thanks. I need a tag to test it in CI.

jluethi:
> Input types are a key-value object where keys are the type string and values are boolean. When we filter for input_types, what is the expectation? Are we filtering for the existence of the key, or are we filtering for a specific boolean value?

Yeah, I'd see this as a potential future place to optimize. Preferably, we'd search for either True or false separately. But let's first keep an eye on how this evolves. I can also imagine we'll have more structured metadata on the task side for this (e.g. one may want to know whether a task supports 2D data, not whether it's 2D exclusive)

PR #671, opened 2024-12-09
Created tasks-list module; moved jschema to components module

### LINKED ISSUE #666: Extract task-list page into separate component, to be re-used in public task-list landing page
(ref https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/871)

The current version https://tcompa.github.io/test-table-github-page example is based on https://datatables.net. We could customize it at will, but this kind of work would be very similar to the new task-list page (ref https://github.com/fractal-analytics-platform/fractal-web/issues/654#issuecomment-2511779208).

With @zonia3000 we are planning to proceed the same way as we did when we isolated a `jschema` component and re-used it for the static sandbox pages in the docs. In this case it would be e.g. a `tasks_table` component, which would be used (with different data) in a public task-list landing page.

Remarks:

1. This approach will be much simpler than re-implementing the same table twice (once for fractal-web and once for the public page), which clearly matters if we move to more complex information.
2. With this approach, we are assuming that we will only have a single format for underlying table data, namely the format of the API response. The data-retrieval procedure for the public page will always have to align with this format.
3. Any update of the table visualization will apply both to the web client and to the task-list public page (as soon as they are updated to the same version). The advantage is that we need to design changes only once, while the disadvantage is that we may have a bit of friction if we have very specific requirements for the public landing page which are not relevant for the Fractal instance. My first guess is that the disadvantage will not matter for a while.
4. The current decisions does not affect the question about where/how we will display this table in the docs. At first, it can be a standalone page (the same as the current sandbox pages). Later we can decide whether it needs to be integrated e.g. as part of https://fractal-analytics-platform.github.io.

---

cc @jluethi - does this sound reasonable?
If so, we'll proceed as in:
- In fractal-web, create the new independent component
- We move from my own testing repo to a Fractal repo - could it be `fractal-task-list` and https://fractal-analytics-platform.github.io/fractal-task-list?
- We setup the two parts of this new repo: data retrieval that matches with the fractal-server API, and static page building based on the new fractal-web component.

---

Out of scope: web components (ref fractal-analytics-platform/fractal-web#512)

zonia3000:
## Checklist before merging

Closes #666


zonia3000:
Regarding the warning:

> ▲ [WARNING] Cannot find base config file "./.svelte-kit/tsconfig.json" [tsconfig.json]
> 
>     ../jsconfig.json:2:12:
>       2 │   "extends": "./.svelte-kit/tsconfig.json",
>         ╵              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

**TL;DR**: It disappears if you run `npm install` on the parent folder.

The build detects that there is a `jsconfig.json` file in the parent folder, so it understands that is currently building a subproject. It then checks if there is a file `.svelte-kit/tsconfig.json` in the parent folder and doesn't find it, so the warning is displayed. To remove the warning we could run `npm install` on the parent folder before executing the subproject build, because running `npm install` also generates the `.svelte-kit` folder.

However, as stated by [a comment on related issue](https://www.github.com/sveltejs/kit/issues/5390#issuecomment-2502040792):

> the warning is supposed to be harmless. It indicates that whatever build system you use (vercel, netlify, multistage docker builds, ...) didn't populate the .svelte-kit folder before invoking vite build.

So, since we don't really need to run `npm install` on the parent folder, I would keep our CI steps as they are, to avoid running an extra step only for removing the warning.

Issue #832, opened 2025-09-15
Display executor errors for Fractal jobs


tcompa:
Starting from fractal-server=2.15.10a2, Fractal jobs have a new property `executor_error_log` (a nullable string).
Here is an example of a response of `GET example.org/api/v2/project/111/latest-job?workflow_id=111&dataset_id=111`:

```json
{
	"id": 111,
	"project_id": 111,
	"project_dump": {
		"id": 111,
		"name": "test-oom",
		"timestamp_created": "2025-09-15T08:12:05.980418Z"
	},
	"user_email": "redacted",
	"slurm_account": null,
	"workflow_id": 111,
	"workflow_dump": {
		"id": 111,
		"name": "oom-1",
		"project_id": 111,
		"timestamp_created": "2025-09-15T08:12:12.556364Z"
	},
	"dataset_id": 111,
	"dataset_dump": {
		"id": 111,
		"name": "oom2",
		"project_id": 111,
		"timestamp_created": "2025-09-15T08:56:37.874945Z",
		"zarr_dir": "redacted"
	},
	"start_timestamp": "2025-09-15T08:56:43.122208+00:00",
	"end_timestamp": "2025-09-15T08:57:06.095000+00:00",
	"status": "failed",
	"log": "JOB ERROR in Fractal job 111:\nTRACEBACK:\nJobExecutionError\n\nAn error occurred.\nOriginal error:\nTraceback (most recent call last):\n  File \"redacted\", line 49, in call_command_wrapper\n    raise TaskExecutionError(\n    ...<2 lines>...\n    )\nfractal_server.app.runner.exceptions.TaskExecutionError: Task failed with returncode=-9.\nSTDERR: \n\n\n",
    # SEE NEXT LINE
	"executor_error_log": "slurmstepd: error: Detected 1 oom_kill event in StepId=111.0. Some of the step tasks have been OOM Killed.\nsrun: error: u20-cva0000-113: task 0: Out Of Memory",
	"working_dir": "redacted",
	"working_dir_user": "redacted",
	"first_task_index": 0,
	"last_task_index": 0,
	"worker_init": null,
	"attribute_filters": {},
	"type_filters": {}
}
```

# To do

## Workflow page

The main change concerns the workflow page

<img width="2573" height="862" alt="Image" src="https://github.com/user-attachments/assets/a6ea2fbb-6b84-4a53-a4c5-f6d7e54fc8ad" />

I suggest we start with the following logic:
[...]

PR #843, opened 2025-10-17
more slurm-error keywords

### LINKED ISSUE #840: Add keywords to SLURM error detection
Three more patterns should be considered in the list of non-false-positive SLURM errors for jobs:
 
* `Application launch failed`
* `Job credential expired`
* `Job step aborted`

tcompa:
close #840

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 *Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web repository.*
 
-# Unreleased
+# 1.21.1
 
+* Added more keywords to match SLURM errors (\#843).
 * Fixed bug in sandbox pages (\#839);
 * Displayed all kinds of recent task activities in tasks management page. (\#835);
 


zonia3000 (approved):

tcompa:
> Was there a specific reason that we're going with the include list? The UX downside is that the first time we (/a user) hits a new slurm error, it's still hidden. Until we report it and add it to the list of errors.
> 
> If we show many false-positives otherwise, that's obviously not helpful either, so the include list has its benefits 😉

The discussion started at https://github.com/fractal-analytics-platform/fractal-web/issues/832#issuecomment-3291381316. I don't think there is a compelling reason for either an include or exclude list.

The same UX downside applies to both cases. With an exclude list, we may mark a stderr content as relevant even if it only a log of SLURM-scheduling events taking longer than usual.

High-level: if these lists change often, then we probably need to review this feature (backend/frontend-side), e.g. through a stronger SLURM integration (including -but not limited to- https://github.com/fractal-analytics-platform/fractal-server/issues/2836). But the assumption here is that we will only ever hit a handful of error categories.

PR #975, opened 2026-02-20
Workflow Templates

### LINKED ISSUE #960: Workflow-template UI - placeholder
Ref https://github.com/fractal-analytics-platform/fractal-server/issues/3186

New features:

## Template-table and CRUD

This feature is re-used in three different places (and it will eventually be a standalone component).

### Top-level templates page, with CRUD

- [x] This is a paginated table of all templates I have access to.
- [x] At the top of the page, there are filtering/sorting options.
- [x] Columns are:
    * user email
    * name
    * version (with no aggregation - for the moment)
- [x] User email is selected from a drop-down of available options (or "Select.." if it is unset).
- The "actions" column includes at least:
    - [x] "info" (which shows a new modal, with at least the name/version/user/group/description/timestamp)
    - [x] "edit" (-> new modal that lets you change description and/or user-group -- see task-group edit modal)
    - [x] "delete" (with confirmation dialog before the API call)
    - [x] "download"
- [x] Somewhere (on top-right?) there should be an "import from file" button, that calls `POST /api/v2/workflow_template/import/`.
- [x] `template_id` query parameter should also be a query parameter for this fractal-web page (e.g. as in example.org/v2/templates/?template_id=123). TBD whether other query parameters for the table should also be re-exposed in the URL.

### Workflow import from template

- [x] From the single-project page, close to the place where we can import a workflow from disk, we should also expose a "import from template" feature.
- [x] This will load the list of available templates (in a modal), with all its filtering/sorting features, and let the user select a single entry.
- [x] In the same way as import from file, it should be possible to edit the name of the workflow to be created - to avoid conflicts.
- [x] The "actions" column does not include edit/delete, but only "select".

### Documentation page

[placeholder] This is similar to the tables above, with some small changes in the "actions" column, in the pagination (page size should be infinite or equal to the total size), some missing columns (user email, version, id, user_group_id).

## Workflow page

- [x] Create template based on the current workflow, from the single-workflow page.
    - [x] Pre-populate template with workflow name.
    - [x] If the current workflow is based on a template I do not own, then pre-populate version with 1.
    - [x] If the current workflow is based on a template I own, then pre-populate version with (current_version+1)
- [x] Display whether the workflow comes from a template - linking to it through example.org/v2/templates/?template_id=123

### LINKED ISSUE #989: Template buttons on single workflow page
Following https://github.com/fractal-analytics-platform/fractal-web/issues/960, the new single-workflow page has two new buttons, called temporarily "Create template" and "Original template":

Image

The page is becoming quite crowded with buttons.

@jluethi is that ok? Or do you have any alternative suggestions?

### LINKED ISSUE #999: Flexible-import fails when importing from template

### LINKED ISSUE #1000: Expose user-group selection in all template-creation modals
This applies to both "import from JSON" and "create from workflow"

The default group: private template

### LINKED ISSUE #1001: Pre-populate template description based on workflow description
(when creating a template based on a workflow)

Priority list:
1. If the workflow comes from a template which has a non-empty description, this is the priority.
2. Next priority is the workflow description, if non-empty
3. Last case is just an empty field

ychiucco:
closes #960 
closes #989
closes #999
closes #1000
closes #1001

## Checklist before merging
- [ ] I added an appropriate entry to `CHANGELOG.md`

### DIFF
--- a/__tests__/v2/workflow_page.test.js
+++ b/__tests__/v2/workflow_page.test.js
@@ -82,6 +82,8 @@ describe('Workflow page', () => {
 							};
 						case '/api/auth/current-user':
 							return { slurm_accounts: [] };
+						case '/api/auth/group':
+							return [];
 						case '/api/v2/project/1/workflow/1/version-update-candidates':
 							return [];
 						default:


tcompa:
Review this modal as already discussed (larger description field, no full-capital "EDIT", and a dropdown-based choice of user groups the same way as for task groups):

image

tcompa:
> Review this modal as already discussed (larger description field, no full-capital "EDIT", and a dropdown-based choice of user groups the same way as for task groups):

Also: use template name rather than ID in the title

tcompa:
Include (a slug of) the template name in the filename, when downloading a JSON file from http://localhost:5173/v2/templates

PR #1007, opened 2026-03-17
Ihb develop

dominique-roux:
## Checklist before merging


zonia3000:
Hi, in case both OAuth2 and the username/password form are present, the login invite text will be duplicated in both the sections. Is this what we want?

image

I see 3 possible alternatives:

* keep the duplication
* define 2 different variables for the 2 kinds of credentials
* display the text only once 

@tcompa what do you think?

PR #1023, opened 2026-03-20
Add "Use pixi.lock file" option in Pixi task collection

### LINKED ISSUE #1011: Expose new form-data property for `POST /api/v2/task/collect/pixi/`
The new parameter is a boolean flag, to be added somewhere in 
Image

Its backend-side default value is `True`, and the frontend-side default should also be set to true.

The parameter name is `use_pixi_lockfile`, and its displayed name should be something like "Use `pixi.lock` file".

This is currently available in https://github.com/fractal-analytics-platform/fractal-server/tree/3216-enable-usage-of-lockfile-in-pixi-task-collection, which is ready to merge (backend-side).

zonia3000:
## Checklist before merging

Closes #1011

### DIFF
--- a/__tests__/v2/PixiTask.test.js
+++ b/__tests__/v2/PixiTask.test.js
@@ -1,0 +1,67 @@
+/// <reference path="../../vitest-setup.d.ts" />
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen } from '@testing-library/svelte';
+import { mockUser } from '../mock/mock-types';
+import userEvent from '@testing-library/user-event';
+
+// Mocking fetch
+global.fetch = vi.fn();
+
+const mockedUser = mockUser({ group_ids_names: [[1, 'All']] });
+
+// The component to be tested must be imported after the mock setup
+import PixiTask from '../../src/lib/components/v2/tasks/PixiTask.svelte';
+
+describe('PixiTask', () => {
+	it('Public task with lockfile and custom pixi version', async () => {
+		const user = userEvent.setup();
+
+		render(PixiTask, {
+			props: { user: mockedUser, defaultGroupName: 'All' }
+		});
+
+		/** @type {import('vitest').Mock} */ (fetch).mockResolvedValue({
+			ok: true,
+			json: () => { }
+		});
+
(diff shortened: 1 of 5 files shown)


tcompa:
In case it's useful for tests, here are two tar.gz archives with/without `pixi.lock`:
* https://github.com/fractal-analytics-platform/fractal-server/blob/3216-enable-usage-of-lockfile-in-pixi-task-collection/tests/data/pixi_tasks/with_lock/mock_pixi_tasks-1.2.3.tar.gz
* https://github.com/fractal-analytics-platform/fractal-server/blob/3216-enable-usage-of-lockfile-in-pixi-task-collection/tests/data/pixi_tasks/no_lock/mock_pixi_tasks-1.2.3.tar.gz


and here is how this is tested backend-side: https://github.com/fractal-analytics-platform/fractal-server/blob/9a3cb0bca30d1045258a7f79d29addeca660ab0b/tests/v2/test_06_tasks_lifecycle/test_api_task_lifecycle_pixi_local.py#L256-L301

tcompa:
> > Release 0.5.0 (https://github.com/fractal-analytics-platform/testing-tasks-mock-pixi/releases/tag/0.5.0) has the `pixi.lock` file in the tar.gz. In principle you should be able to use this one for testing - let me know if anything goes wrong.
> 
> I tried this but it takes too long to install (around 3 minutes) and it is not suitable to run in end to end tests. We could increase test timeout, but I would prefer to have quicker tests. Let's stick on 0.2.2 for a while. As an alternative, I can also test this feature using only unit tests.

I see - it's most likely due to https://github.com/fractal-analytics-platform/testing-tasks-mock-pixi/commit/24ed32fc5711e117b5a5899383945dd1927270ae.

cc @lorenzocerrone: are you relying on https://github.com/fractal-analytics-platform/testing-tasks-mock-pixi anywhere? If not, I will move it back to a vanilla project (without any torch) so that it can be used in CI.

lorenzocerrone:
> cc @lorenzocerrone: are you relying on https://github.com/fractal-analytics-platform/testing-tasks-mock-pixi anywhere? If not, I will move it back to a vanilla project (without any torch) so that it can be used in CI.

No, feel free to simplify it as much as you want.

PR #1041, opened 2026-03-27
Fix groups visibility for standard users on template creation

ychiucco:
## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 # 1.24.1 (unreleased)
 
-* Fix user group visibility for standard user on templates creation (\#1041).
+* Add `FRACTAL_HIDE_BASIC_AUTH` environment variable (\#1039).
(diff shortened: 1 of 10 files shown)


src/lib/components/v2/templates/TemplateInfoModal.svelte
```diff
@@ -38,14 +36,18 @@
...
 				);
 				if (response2.ok) {
-					group = await response2.json()
+					const res = await response2.json()
+					groupName = res.group_ids_names.find(([id]) => id === template?.user_group_id)?.[1] ?? undefined;
+					if (!groupName) {
+						throw new Error(`User group ${template?.user_group_id} not found`);
+					}
 				} else {
 					await formErrorHandler.handleErrorResponse(response2);
```
zonia3000: Here you can use the following:
```js
errorAlert = displayStandardErrorAlert(
	await getAlertErrorFromResponse(response),
	'errorAlert-infoTemplateModal'
);
```

PR #1052, opened 2026-04-02
UI for project ownership tranfer

### LINKED ISSUE #1049: UI for project-ownership tranfer
Ref:
* https://github.com/fractal-analytics-platform/fractal-server/issues/3218
* https://github.com/fractal-analytics-platform/fractal-server/pull/3261

The first plan for this would be as follows

1. New admin page "Projects"
    - Same structure as `v2/admin/tasks`
    - Based on `GET /admin/v2/project/?user_email={user_email}&project_id={project_id}?name={name_part}` (+pagination)
2. In the "Options" column, a single transfer-ownership button, that opens a modal (for a given `project_id`).
    - The main message should read something like "Transfer the ownership of project {project.id} ({project.name}) to a new user. WARNING: The current owner ({current_owner_email}) will not have access to this project any more.".
    - A dropdown menu with all users' emails should let you choose `user_id`
    - The modal eventually makes a call to `PATCH /admin/v2/project/{project_id}/?user_id={user_id}` (I think we first need a confirmation prompt)
    - API errors must be handled
    - Before confirmation, compute the profile IDs for both users and compare them. If they do not match, display an additional warning ("WARNING: {user_old} profile ({profile_old.name}) differs from {user_new} profile ({profile_new.name}. The new project owner may not be able to run jobs on this projects, due to on-disk permissions."))

ychiucco:
close #1049 

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ _Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web rep
 
 - Add support for conditionals in JSON Schemas (`dependentRequired`, `if`/`then`/`else`) (\#1050).
 - Apply prettier and fix static code analysis CI configuration (\#1051).
+- Add new admin page for projects (\#1052).
 
 # 1.25.1
 
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -96,8 +96,7 @@ export default defineConfig({
 
 	webServer: [
 		{
-			command:
-				'./tests/start-test-server.sh --branch 3218-expose-transfer-project-ownership-endpoint-for-admins-only',
+			command: './tests/start-test-server.sh --branch main',
 			port: 8000,
 			waitForPort: true,
 			stdout: 'pipe',


src/routes/v2/admin/projects/+page.svelte
```diff
@@ -0,0 +1,314 @@
...
+		id = '';
+		userId = undefined;
+		searched = false;
+		results = undefined;
+		currentPage = 1;
+		totalCount = 0;
+		pageSize = 50;
+	}
+
+	async function handleChangeOwner() {
```
zonia3000: Here you can use the modal `confirmAndHide` method. The error message div should have the proper id.

```js
changeOwnershipModal?.confirmAndHide(async () => {
  const response = await fetch(
    `/api/admin/v2/project/${selectedProject?.id}?user_id=${newOwnerId}`,
    {
      method: 'PATCH',
      credentials: 'include'
    }
  );
  if (!response.ok) {
    throw await getAlertErrorFromResponse(response);
  }
  await searchProject(currentPage, pageSize);
});
```

PR #1071, opened 2026-04-15
Use slim=true instead of args_schema=false when loading task groups

### LINKED ISSUE #1026: Review response body/schema in `GET /api/v2/task-group/`
In https://github.com/fractal-analytics-platform/fractal-server/issues/3239 we had a look on how to optimize `GET /api/v2/task-group/`, notably for when it is used in

Image

One of the most straightforward optimization would be to reduce the response-body size - and indeed we are already including `only_active=true&args_schema=false`.

Question: out of all the https://github.com/fractal-analytics-platform/fractal-server/blob/901ce7e6b86bb840c98a9ae405c41e7ef49a7353/fractal_server/app/schemas/v2/task.py#L93-L116 properties, which ones are actually used in this context?

The screenshot above suggests we mostly need:
* id
* category
* modality
* authors, tags
* version

Is this correct?

If so, we could create a way to get a smaller response body [a new query parameter (e.g. `?slim=true`), or a new response schema, or a new endpoint.. TBD] and review how sizeable the effect would be.

zonia3000:
## Checklist before merging

Closes #1026

### DIFF
--- a/src/lib/server/api/v2/task_api.js
+++ b/src/lib/server/api/v2/task_api.js
@@ -7,15 +7,15 @@ const logger = getLogger('task API [v2]');
 /**
  * Fetches a list of task groups from the server
  * @param {typeof fetch} fetch
- * @param {boolean|false=} loadArgsSchema
+ * @param {boolean|false=} slim
  * @param {boolean|false=} onlyActive
  * @returns {Promise<Array<[ string, Array<import('fractal-components/types/api').TaskGroupV2> ]>>}
  */
-export async function listTaskGroups(fetch, loadArgsSchema = false, onlyActive = false) {
+export async function listTaskGroups(fetch, slim = false, onlyActive = false) {
 	logger.debug('Fetching task groups');
 
 	const response = await fetch(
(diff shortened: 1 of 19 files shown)


src/lib/server/api/v2/task_api.js
```diff
@@ -7,15 +7,15 @@ const logger = getLogger('task API [v2]');
 /**
  * Fetches a list of task groups from the server
  * @param {typeof fetch} fetch
- * @param {boolean|false=} loadArgsSchema
+ * @param {boolean|false=} slim
  * @param {boolean|false=} onlyActive
  * @returns {Promise<Array<[ string, Array<import('fractal-components/types/api').TaskGroupV2> ]>>}
  */
-export async function listTaskGroups(fetch, loadArgsSchema = false, onlyActive = false) {
```
tcompa, line 14: Quick question: the default is changing from `loadArgsSchema=false` (which returns the objects without args schemas) to `slim=false` (which returns the full objects). Is this intentional?

zonia3000: Yes, because we are overriding the default in the caller function and it seemed more appropriate to keep this default as the backend default.

tcompa: Makes sense - thanks

PR #1072, opened 2026-04-15
Visualize template description from create-workflow modal

### LINKED ISSUE #1044: Make workflow template info (e.g. description) accessible from `Create new workflow` modal
When creating new workflow from a template, I'm presented with this modal:

Image

While writing up the template announcement, an additional UI requirement came up for me: We should make the template info, especially workflow description, accessible from this page.

This has 2 elements:
1. Make workflow info at all accessible (e.g. with an "i" button) => double modal issue? => pop-up box like in task addition screen?
2. Give workflow description enough prominence so that users find it

Point 2 is especially relevant, given that titles alone aren't always descriptive enough. If we can show a modal, we could show something similar to:

Image

But both here as well as on the workflow template page, it would be good to reorder the info to be more user optimised. Proposed order:
1. Name
2. Description
3. User email
4. Version
5. User group
6. Creation timestamp
7. Last use timestamp
8. Template ID

If we cannot show a modal on the the create page, showing an "i" with a click pop-up effect like on the task addition page actually would be even better here:

Image

We can limit the information shown in this hover to just the description.

ychiucco:
closes #1044 

## Checklist before merging


src/lib/components/v2/templates/TemplatesTable.svelte
```diff
@@ -373,13 +376,56 @@
...
-								<td class="col-5">{templateGroup.template_name}</td>
+
+								{#if modalType === 'select'}
+									<td class="col-5">
+										<div class="d-flex align-items-center gap-2">
+											<span>{templateGroup.template_name}</span>
+											<button
+												class="btn btn-link p-0"
+												onclick={async () => {
+													if (!descriptions[index]) {
```
zonia3000: Can we move this to a dedicated function in the `<script>` section? Better to not put too many lines of JavaScript in the template part.

zonia3000 (approved):

PR #1076, opened 2026-04-16
Support creating templates loading workflow json file

ychiucco:
closes #1062 

## Checklist before merging

### DIFF
--- a/src/lib/components/v2/templates/TemplateImportModal.svelte
+++ b/src/lib/components/v2/templates/TemplateImportModal.svelte
@@ -21,15 +21,22 @@
 	let fileInput = $state(undefined);
 
 	/**
+	 * @typedef {import('fractal-components/types/api').WorkflowImport} WorkflowImport
+	 * @typedef {import('fractal-components/types/api').WorkflowTemplateImport} WorkflowTemplateImport
+	 */
+
+	/**
 	 * @typedef {Object} WorkflowTemplateImportTemporary
 	 * @property {string|null} name
 	 * @property {number|null} version
 	 * @property {string|null} description
-	 * @property {import('fractal-components/types/api').WorkflowImport} data
+	 * @property {WorkflowImport} data
 	 */
 
 	/** @type {WorkflowTemplateImportTemporary|undefined} */
 	let template = $state(undefined);
+	/** @type {string|undefined} */
+	let errorMessage = $state(undefined);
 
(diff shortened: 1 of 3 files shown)


src/lib/components/v2/templates/TemplateImportModal.svelte
```diff
@@ -60,27 +75,62 @@
...
-		}
-
-		if (templateName) {
-			console.log(`Overriding template name from '${template.name}' to '${templateName}'`);
-			template.name = templateName;
-		}
-		if (templateVersion) {
-			console.log(`Overriding template version from '${template.version}' to '${templateVersion}'`);
-			template.version = templateVersion;
+			throw new AlertError('The imported file must be a Workflow or a WorkflowTemplate');
```
tcompa: Quick question: is this line reachable? When do we reach this one, and when do we reach
```js
            errorMessage = 'the input file is not a Workflow nor a WorkflowTemplate.';
			return;
```
?

src/lib/components/v2/templates/TemplateImportModal.svelte
```diff
@@ -60,27 +74,57 @@
...
-		/** @type {import('fractal-components/types/api').WorkflowTemplateImport} */
-		let template;
-
-		try {
-			template = JSON.parse(await templateFile.text());
-		} catch (err) {
-			console.error(err);
-			throw new AlertError('The workflow file is not a valid JSON file');
-		}
+		const input = JSON.parse(await templateFile.text());
```
zonia3000, line 108: Add a try/catch to handle the case of a user selecting a file containing invalid JSON data.

src/lib/components/v2/templates/TemplateImportModal.svelte
```diff
@@ -137,49 +176,80 @@
...
+						{#each groups as group, index (index)}
+							<option value={group[0]}>{group[1]}</option>
+						{/each}
+					</select>
+				</div>
+
+				<button
+					class="btn btn-primary mt-2"
+					disabled={creating || !template.name || !template.version || template.version < 1}
+					aria-label="Import template"
```
zonia3000, line 239: Just a note: `aria-label` is useful when the element doesn't contain any text (e.g. a button using just an icon). In this case it is not wrong to use the aria-label, but it is redundant.

zonia3000 (approved):

PR #1079, opened 2026-04-20
New UI for `GET admin/v2/users` csv

### LINKED ISSUE #1070: New UI for `GET /admin/v2/users-csv/`
Developed in https://github.com/fractal-analytics-platform/fractal-server/pull/3276.

1. This could be e.g. a "Users info" page in the admin area.
2. The endpoint takes three query parameters:
    * `exclude_zero_jobs`, a boolean
    * `start_timestamp_min` a timezone-aware timestamp
    * `start_timestamp_max` a timezone-aware timestamp
3. The UI should expose two elements:
    * A component to set `exclude_zero_jobs` (false by default), with description "Exclude users with no jobs in the selected time interval".
    * A component where the user can pick a month+year combination, or nothing (*).
        * The description should read "Time interval for users' jobs"
        * If nothing is selected, `start_timestamp_min` and `start_timestamp_max` are either set to `null` or not included in the request body.
        * If April 2025 is selected, `start_timestamp_min` is set to 00:00:01 of April 1st and `start_timestamp_max` is set to 00:00:01 of May 1st (or anything similar - meaning that the interval should fully cover all days of the selected month).
4. The response has type `text/csv` and a `"Content-Disposition": "attachment; filename=users.csv"` header. 
5. Clicking "submit" should prompt the user to download the file.
6. I'd replace the default filename client-side (if possible), with:
    * `users-all-time.csv` when no month+year was selected
    * `users-2025-04.csv` when April 2025 is selected

(*) Note that the API is more flexible than this, as it exposes an arbitrary time interval. For the moment we should keep the web-client a simpler, and only expose the month+year option.

ychiucco:
closes #1070 

## Checklist before merging


src/routes/v2/admin/users-info/+page.svelte
```diff
@@ -0,0 +1,133 @@
...
+		const downloadUrl = URL.createObjectURL(blob);
+
+		const a = document.createElement('a');
+		a.href = downloadUrl;
+		a.download = buildFilename();
+		document.body.appendChild(a);
+		a.click();
+
+		a.remove();
+		URL.revokeObjectURL(downloadUrl);
```
tcompa: I am a bit confused by the pattern of first fetching and then exposing the download-blob feature. Could we skip the `fetch` call, as we do in the snippet below?

https://github.com/fractal-analytics-platform/fractal-web/blob/63565b56c69fbcab8ff8627adbf1f30b8df75faa/src/lib/components/v2/jobs/JobsList.svelte#L574-L581

zonia3000: Yes, the download-blob feature is useful mostly when the download data needs to be generated or manipulated from the frontend. Notice however that, since we have dynamic parameters, to generate a dynamic link we need to populate it using `$effect` or using a function with all the parameters, otherwise Svelte reactivity will not be triggered.

src/routes/v2/admin/users-info/+page.svelte
```diff
@@ -0,0 +1,133 @@
...
+	<h1 class="fw-light mb-4">Download users info</h1>
+
+	<div class="container mt-3">
+		<div class="row mt-3">
+			<div class="col-md-3 col-lg-3 mt-2">Restrict to jobs started on:</div>
+			<div class="col-md-9 col-lg-9">
+				<div class="row row-cols-md-auto">
+					<div class="col-12 mt-1">
+						<select
+							class="form-select {!startMonth && startYear ? 'is-invalid' : ''}"
```
zonia3000: If you want you can use also this format:

```
class="form-select"
class:is-invalid={startMonth && !startYear}
```

It is usually more readable, especially when there are many dynamic classes for different conditions

src/routes/v2/admin/users-info/+page.svelte
```diff
@@ -0,0 +1,133 @@
...
+				</div>
+			</div>
+		</div>
+
+		<div class="row mt-3">
+			<div class="col-md-3 col-lg-3 mt-2">Exclude users with no jobs:</div>
+			<div class="col-md-9 col-lg-9">
+				<div class="row row-cols-md-auto">
+					<div class="col-12 mt-1">
+						<input type="checkbox" class="form-check-input" bind:checked={excludeZeroJobs} />
```
zonia3000: I would set "Exclude users with no jobs" as `label` instead of `div` and reference it here, otherwise we would need an aria-label.

zonia3000 (approved):

Issue #1086, opened 2026-04-22
UI for `dataset.tags`


tcompa:
See https://github.com/fractal-analytics-platform/fractal-server/issues/3217

- [ ] In the create-dataset modal, let the user select one or multiple tags. Tags can be defined in two coexisting ways:
    * By picking them from a dropdown menu, populated by fetching `GET /api/v2/project/{project_id}/dataset-tag/`.
    * By typing them manually
- [ ] In the "Info" modal of the single-dataset page, display the current tags.
- [ ] In the "Info" modal of the single-dataset page, let the user edit the dataset tags. This has the same structure as the create-dataset one: the user can either use existing tags, or write new ones.
- [ ] In the dataset-list table of the single-project page, add a column with tags.
- [ ] In the dataset-list table of the single-project page, add a filtering option which works fully client-side.
    * A dropdown menu should list all project tags (either via `GET /api/v2/project/{project_id}/dataset-tag/`, or by directly parsing the table contents)
    * The user should be able to select zero, one or multiple tags in the menu.
    * Only dataset that have all the selected tags should be displayed.
    * There should be no "Apply" button, since this is a client-side filtering.
    * There should be a "Reset" button (which should only be active when at least one tag is selected), that removes any filter.

tcompa:
This requires a better specification of https://github.com/fractal-analytics-platform/fractal-server/issues/3217, which is still ongoing.

Closing.

PR #1089, opened 2026-04-23
Datasets page

### LINKED ISSUE #1087: UI for datasets page
Ref https://github.com/fractal-analytics-platform/fractal-server/issues/3293

ychiucco:
closes #1087

## Checklist before merging

### DIFF
--- a/src/routes/v2/datasets/+page.svelte
+++ b/src/routes/v2/datasets/+page.svelte
@@ -56,7 +56,7 @@
 	const isDirtyFromApplied = $derived(
 		currentState.projectName !== lastAppliedState.projectName ||
 			currentState.onlyOwned !== lastAppliedState.onlyOwned ||
-			currentState.datasetName != lastAppliedState.datasetName
+			currentState.datasetName !== lastAppliedState.datasetName
 	);
 
 	const applyClass = $derived(isDirtyFromApplied ? 'btn-primary' : 'btn-secondary');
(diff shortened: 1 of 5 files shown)


src/routes/v2/datasets/+page.svelte
```diff
@@ -0,0 +1,240 @@
...
+
+	onMount(async () => {
+		await searchDatasets();
+	});
+
+	async function searchDatasets() {
+		const url = new URL(window.location.href);
+		// Headers
+		const headers = new Headers();
+		headers.set('Content-Type', 'application/json');
```
zonia3000: No need for this header in a GET request

src/routes/v2/datasets/+page.svelte
```diff
@@ -0,0 +1,240 @@
...
+
+		let response = await fetch(`/api/v2/dataset?${params.toString()}`, {
+			method: 'GET',
+			headers,
+			credentials: 'include'
+		});
+		if (response.ok) {
+			datasetPage = await response.json();
+		} else {
+			throw await getAlertErrorFromResponse(response);
```
zonia3000: This exception is never caught. You can use the approach with `errorAlert = displayStandardErrorAlert(...`.
In similar cases, where we don't have an easy way to trigger validation errors, I usually do a manual tests for errors by temporarily changing the URL to a non-existent one, just to see if the error message is displayed on the page.

zonia3000:
More general question: from which link is the user supposed to reach this new page?

tcompa:
> More general question: from which link is the user supposed to reach this new page?

It will be a "Datasets" link from the top menu, but for the moment we are keeping it "hidden" (that is, only accessible via direct URL) for further testing. The major missing part is https://github.com/fractal-analytics-platform/fractal-web/issues/1086, which still requires some discussions.

zonia3000 (approved):

PR #1103, opened 2026-05-12
Use new combined latest-job & status endpoint

### LINKED ISSUE #758: [requires backend updates] Simplify status logic (after combining latest-job and status endpoints)
Related to https://github.com/fractal-analytics-platform/fractal-web/issues/755.

If we had a single backend endpoint with a response like:
```json
{
  "job_status": "submitted",
  "statuses": {
    "65": "done",
    "66": "submitted"
  }
}
```
this would simplify the internal logic for which calls to make in background to update the workflow-task status.

Note: it'd be simpler to do this after we deprecate the status-legacy endpoint.

### LINKED ISSUE #1101: Drop obsolete task-group properties
Image

zonia3000:
## Checklist before merging

Close #758
Close #1101

### DIFF
--- a/components/src/lib/types/api.d.ts
+++ b/components/src/lib/types/api.d.ts
@@ -232,6 +232,10 @@ export type ImagesStatus =
 			num_available_images: number | null;
 	  };
 
+export type JobWithTaskStatuses = ApplyWorkflowV2 & {
+	task_statuses: Record<number, ImagesStatus>;
+};
+
 export type WorkflowV2 = {
 	id: number;
 	name: string;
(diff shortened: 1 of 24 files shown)


components/src/lib/types/api.d.ts
```diff
@@ -232,6 +232,10 @@ export type ImagesStatus =
 			num_available_images: number | null;
 	  };
 
+export type JobWithTaskStatuses = ApplyWorkflowV2 & {
```
tcompa: Should we update this from the obsolete `ApplyWorkflowV2` into e.g. `JobV2`?

PR #1104, opened 2026-05-12
UI for task-group reset

### LINKED ISSUE #1102: UI for admin/reset task
On `v2/admin/task-groups`, this modal should include a "Reset task group" option.
Image

---

This leads to `v2/admin/task-groups/{id}/reset` - which is somewhat similar to the top part of `v2/tasks/management`.

Available tabs/options: PyPI, Local whl, Pixi

### UI for PyPI case

Endpoint to be called `POST /admin/v2/task-group/{id}/reset/pip/`

PyPI endpoint takes a request body like
```python
class TaskGroupOverridesPip(BaseModel):
    pip_extras: str | None = None
    python_version: NonEmptyStr | None = None
    pinned_package_versions_pre: Json[dict[NonEmptyStr, NonEmptyStr]] = Field(
        default_factory=dict
    )
    pinned_package_versions_post: Json[dict[NonEmptyStr, NonEmptyStr]] = Field(
        default_factory=dict
    )
```

UI:
* `python_version` --> same as `v2/tasks/management`
* `pip_extras` (displayed name: "Package extras") must be slightly updated, because it should support both `""` and "null` (and `"something"`). Maybe a checkbox for "include / do not include" and only when this is checked we send something.
* "Pinned packages versions:" remains as in the management page.

### UI for wheel-file case

For the moment this would be identical --> we can group them in a single "PyPI / local whl" tab.

### UI for pixi case

Endpoint to be called `POST /admin/v2/task-group/{id}/reset/pixi/`

For the moment we should only include a boolean `use_pixi_lockfile` argument, and nothing else.

## Other aspects

On top of the form, we could display task-group pkg_name, version, and email of the owner. As in

> You are about to reset task-group {pkg_name} {version}, collected by {user_email}.

Clicking requires a confirmation dialog, mentioning that this is non-reversible.

Successful POST response redirects to the task-group activity page (`/v2/admin/task-groups/activities?activity_id={some-id}`)

ychiucco:
closes #1102 

## Checklist before merging


src/lib/components/v2/tasks/TaskGroupManageModal.svelte
```diff
@@ -87,8 +88,8 @@
 		if (response.ok) {
 			const result = await response.json();
 			saving = false;
-			const page = admin ? `/v2/admin/task-groups/activities` : `/v2/tasks/activities`;
-			goto(`${page}?activity_id=${result.id}`);
+			const url = admin ? `/v2/admin/task-groups/activities` : `/v2/tasks/activities`;
+			goto(`${url}?activity_id=${result.id}`);
```
zonia3000: `await goto`

src/lib/components/v2/tasks/TaskGroupManageModal.svelte
```diff
@@ -172,6 +173,16 @@
...
 						</button>
+						{#if page.data.userInfo.is_superuser}
+							<button
+								class="btn btn-primary"
+								onclick={async () => {
+									await goto(`/v2/admin/task-groups/${taskGroup?.id}/reset`);
+								}}
+							>
+								Reset task group
+							</button>
```
zonia3000: Here we can use a simple link:
```html
<a href="/v2/admin/task-groups/{taskGroup?.id}/reset" class="btn btn-primary">
  Reset task group
</a>
```

src/routes/v2/admin/task-groups/[taskGroupId]/reset/+page.svelte
```diff
@@ -0,0 +1,330 @@
...
+			headers: {
+				'Content-Type': 'application/json'
+			},
+			body: JSON.stringify(payload)
+		});
+		if (response.ok) {
+			const result = /** @type {import('fractal-components/types/api').TaskGroupActivityV2} */ (
+				await response.json()
+			);
+			goto(`/v2/admin/task-groups/activities?activity_id=${result.id}`);
```
Comment (zonia3000) [inline src/routes/v2/admin/task-groups/[taskGroupId]/reset/+page.svelte]: `await goto`

src/lib/components/v2/tasks/TaskGroupManageModal.svelte
zonia3000, line 12: This components provides the `admin` property to be included both in the standard tasks management page and in the admin task-groups page, in order to behave differently depending on the case.

Do we need to expose the "Reset" button in both the pages? If not, we can probably rely on this property instead of using `page.data.userInfo.is_superuser` in line 176.

src/lib/components/v2/tasks/TaskGroupManageModal.svelte
```diff
@@ -172,6 +173,16 @@
 						<button class="btn btn-primary" onclick={() => handleEditTaskGroup(true)}>
 							Reactivate task group
 						</button>
+						{#if page.data.userInfo.is_superuser}
```
zonia3000: See comment about `let { admin } = $props();` (line 13)

zonia3000 (approved):

PR #1110, opened 2026-05-18
Ajv support for contentMediaType and contentSchema

### LINKED ISSUE #1106: [ON HOLD] Explore support for `contentMediaType` and `contentSchema`
Ref https://github.com/fractal-analytics-platform/example-fractal-task-arguments/issues/20.

We have a new kind of task-arguments JSON schemas which may be relevant, see e.g. this one: https://github.com/fractal-analytics-platform/example-fractal-task-arguments/blob/0d592a8503bf7ed765292f413e90014caa92277d/single_json_schemas/task7_json_strings-parallel.json.

Notice the keywords `contentMediaType` and `contentSchema`.

The JSON Schema above is not interpreted as intended by at least two tools:
* https://fractal-analytics-platform.github.io/fractal-web/sandbox/#jschema
* https://rjsf-team.github.io/react-jsonschema-form/

In both cases, I would expect that passing a string value like `"abc"` would lead to a validation error, but this is not the case.

Notable quote from https://ajv.js.org/json-schema.html#metadata-keywords:

> Ajv does not implement validation of the keywords examples, contentEncoding and contentMediaType but it reserves them. If you want to create a plugin that implements any of them, it should remove these keywords from the instance.

See also https://github.com/json-schema-org/json-schema-spec/issues/646.

I don't see any obvious workaround here apart from writing an ajv plugin (https://ajv.js.org/packages/), which may be very complex (or not.. TBD).

### LINKED ISSUE #1107: [ON HOLD] Customization for large JSON-text task arguments
If there is progress on #1106, we should find out how to best provide possibly long JSON values to a task.

First questions:
* Is there a maximum size for a text input field?
* Can we make _some_ input fields larger (especially in the vertical direction).
* Open question: how would we determine _which_ input fields should be made larger? A starting point, just at the demo level, could be that we have a simple pattern that we identify in the JSON schema - and it is present we switch to the large-field version. But we should review this approach and make it more general/robust.

zonia3000:
## Checklist before merging

- [ ] I added an appropriate entry to `CHANGELOG.md`

Closes #1106
Closes #1107

### DIFF
--- a/components/__tests__/jschema/properties/jsonString.test.js
+++ b/components/__tests__/jschema/properties/jsonString.test.js
@@ -1,0 +1,63 @@
+import { describe, it, expect } from 'vitest';
+import { screen } from '@testing-library/svelte';
+import { renderSchema } from './property_test_utils';
+import userEvent from '@testing-library/user-event';
+
+describe('JSON string properties', () => {
+	it('JSON string', async () => {
+		const user = userEvent.setup();
+		const { component } = renderSchema(
+			{
+				$defs: {
+					JsonChild: {
+						type: 'object',
+						properties: {
+							foo: {
+								type: 'string'
+							},
+							bar: {
+								type: 'number'
+							}
+						},
+						required: ['foo', 'bar']
+					}
+				},
+				type: 'object',
+				properties: {
+					test: {
+						type: 'object',
+						properties: {
+							json_prop: {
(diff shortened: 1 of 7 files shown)


tcompa:
This looks good.

Closing as of https://github.com/fractal-analytics-platform/fractal-web/issues/1106#issuecomment-4496869407.

PR #1149, opened 2026-05-27
Bootstrap and SlimSelect import changes; Security headers configurations

### LINKED ISSUE #1146: Review/include security headers
See e.g. https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#security-headers

zonia3000:
## Checklist before merging

Closes #1146


zonia3000:
Previously we imported Bootstrap from CDN, using a `<script>` tag.

This was mainly because SvelteKit attempts to load the library during SSR, but Bootstrap calls document methods, which are not available server-side, and then this causes errors.

I've found the following technique for Bootstrap at page load:

```svelte
<script>
	import bootstrapScriptSrc from 'bootstrap/dist/js/bootstrap.bundle.js?url';
</script>
<svelte:head>
	<script src={bootstrapScriptSrc}></script>
</svelte:head>
```

In this way we have Bootstrap handled as all the other dependencies (using package.json).

In theory, we could now load all the Bootstrap components using standard imports:

```js
import { Modal } from 'bootstrap';
```

However this is not working because sometimes the components are rendered during SSR, so I didn't touch the Bootstrap components usage in the main code, which is currently like this:

```js
// eslint-disable-next-line no-undef
const bootstrapModal = bootstrap.Modal.getInstance(modalElement);
```

We are using the global variable `bootstrap`, which causes some eslint warnings (which we had already suppressed).

I was able to use these standard imports in sandbox, tasks-list and templates-list subprojects, since those are simple SPA and don't use SSR.

src/routes/auth/login/oauth2/+layout.server.js
```diff
@@ -31,7 +31,7 @@ export async function load({ request, fetch, cookies }) {
 
 	let fastApiToken = null;
 	for (const cookie of receivedCookies) {
-		const match = cookie.match(/fastapiusersauth=([^;]*);/);
+		const match = cookie.match(/fastapiusersauth=([^;]*);?/);
```
tcompa, line 34: Minor question: should we use `fastapiusersauth` or `AUTH_COOKIE_NAME` here?

zonia3000: `AUTH_COOKIE_NAME` is currently for the name used only by Svelte (which can contain the `__Host-` prefix). Do you want that we use the same for the backend?

PR #1150, opened 2026-05-28
Highlight warnings from task logs

### LINKED ISSUE #1148: Highlight "WARNING"s from task logs
fractal-server issue: https://github.com/fractal-analytics-platform/fractal-server/issues/3238/
fractal-server PR: https://github.com/fractal-analytics-platform/fractal-server/pull/3351/

We are looking at the single-workflow page.

### 0 clicks

TBD whether something should appear top-level (see points 3 and 5 of the API section above).

### 1 click

I expand the list of history runs for a given workflow task, and on each history-run row I see a count of warnings.

### 2 clicks

I open the history-units modal for a given history run, and I see which units have warnings and which ones do not.

### 3 clicks

I click on the "Logs" button for one of the units with warnings (perhaps the button itself is highlighted in red).
I see more than one tab, namely "Warnings" (pre-selected) and "Full logs" (similar to what we do with task vs SLURM logs).
The filtering is taking place only here, and client-side.

## Some remarks

* In this solution, the frontend is in charge of creating the "filtered" logs (that is, the one which only include the warnings). This is in principle suboptimal, but such an isssue is strongly mitigated by the fact that this only takes place for one history unit at the time. That is, the frontend only has to process e.g. 1000 log lines, which is not an issue. If we later find out that this is a problem, we can move the pattern-matching to the backend and have a query parameter for the get-logs endpoint that says "only give me warning lines".
* What is TBD and could be possibly improved is to reduce the number of clicks between "I know this task had some warnings" and "I am reading the warnings text".

ychiucco:
closes #1148 

## Checklist before merging


src/lib/components/jobs/RunStatus.svelte
```diff
@@ -22,6 +22,27 @@
 
 <span class="d-flex">
 	<span class="d-flex">
+		{#if run.num_units_with_warnings > 0}
+			<button
+				aria-label="Warnings"
+				aria-disabled="true"
+				class="status-modal-btn btn btn-link text-decoration-none p-0"
+				onclick={(event) => event.preventDefault()}
```
zonia3000: Try using `type="button"` instead of  `onclick={(event) => event.preventDefault()}`.
More info here: https://mtsknn.fi/blog/button-type-button/

tests/v2/workflows/run_mock_tasks.spec.js
```diff
@@ -325,6 +335,27 @@ test('Run mock tasks [v2]', async ({ page, workflow }) => {
...
+		await page.getByLabel('Selector for attribute well', { exact: true }).getByText('All').click();
+		await page.getByRole('option', { name: 'A01' }).click();
+		await page.getByRole('button', { name: 'Apply' }).click();
+		await page.getByRole('button', { name: 'Run', exact: true }).click();
+		await page.getByRole('button', { name: 'Confirm' }).click();
+		await page.getByRole('button', { name: 'Show runs generic_task Done' }).click();
+		await page.getByRole('button', { name: 'Failed images', exact: true }).click();
+
+		const logsButtonRed = await page.getByRole('button', { name: 'Logs' }).nth(0);
+		const logsButtonYellow = await page.getByRole('button', { name: 'Logs' }).nth(1);
```
zonia3000: `nth()` returns a `Locator`, not `Promise<Locator>`, so `await` is not needed here and in the previous line.

zonia3000 (approved):

PR #1180, opened 2026-06-17
Remove usage of GET /api/v2/task/ endpoint

### LINKED ISSUE #1179: Refactor search for health-check task
As far as I can tell, the `getHealthCheckTask` is the only place where `GET /api/v2/task/` is used (in either `fractal-web` or `fractal-client`):
https://github.com/fractal-analytics-platform/fractal-web/blob/4659c1e5478f349104a3f368d3b02086d980d946/src/routes/healthcheck/%2Bpage.svelte#L142-L154

Did I miss any other usage of this endpoint?

---

If we could refactor this function, I would then proceed and fully remove the endpoint (this started as a simplification plan at https://github.com/fractal-analytics-platform/fractal-server/issues/3366).

Alternative options could be:
1. Fetching the whole list of task groups (`GET /api/v2/task-group/`) and looking through it.
2. Trying to create the task (as [here](https://github.com/fractal-analytics-platform/fractal-web/blob/4659c1e5478f349104a3f368d3b02086d980d946/src/routes/healthcheck/%2Bpage.svelte#L121-L133)) and handling a 422 response (but it's TBD whether the client would have enough information to proceed).
3. ... other options?

If this turns out to be too complex, we can obviously leave it in the current state and we'll keep maintaining the backend endpoint.

zonia3000:
## Checklist before merging

Closes #1179


src/routes/healthcheck/+page.svelte
```diff
@@ -140,17 +140,14 @@
...
 	async function getHealthCheckTask() {
-		const response = await fetch(`/api/v2/task?slim=true`, {
-			method: 'GET',
-			credentials: 'include'
-		});
-		if (!response.ok) {
-			throw await getAlertErrorFromResponse(response);
+		const request = await fetch(`/api/v2/task-group?only_active=true&slim=true`);
+		if (!request.ok) {
+			throw await getAlertErrorFromResponse(request);
```
tcompa, line 145: A brief clarification:
Is there a reason we do not include method/credentials parameters any more? Are they simply the default values?

zonia3000: Yes, GET is the default method, so it can be omitted. In the code we are not very coherent in these cases. Sometimes we put it and sometimes not. Maybe we should decide a convention and stick on it.

tcompa: Thanks - but isn't `credentials` changing, from `'include'` (in the previous version) to `'same-origin'` (the [default](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#including_credentials))? Which one is more appropriate?

zonia3000: Yes, we can omit that too and use the default. We also have some `mode: 'cors'` which can also be removed.

tcompa: OK - let's move this to https://github.com/fractal-analytics-platform/fractal-web/issues/1181.

PR #1182, opened 2026-06-18
Improve accessibility and add accessibility tests

### LINKED ISSUE #1181: Make usage of `fetch` defaults homogeneous
E.g.
1. Do we always specify `GET`?
2. What `credentials` option do we use (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#including_credentials)?
3. Review `mode: 'cors'`

Let's write this down somewhere in https://github.com/fractal-analytics-platform/fractal-web/tree/main/docs/development (or in a `CONTRIBUTING.md` file, if it's preferred) and adopt it across the repo.

zonia3000:
## Checklist before merging

Closes #1181


zonia3000:
# Some notes about the accessibility tests

Playwright documentation: https://playwright.dev/docs/accessibility-testing

I added a base test fixture that contains a `checkAccessibility()` method. This method is automatically executed at the end of any test. Notice that it checks the accessibility of the page in that given instant. So, if a test involves opening modals or navigating to different pages, some accessibility issues could not be detected by this base check.

For this reason I added some extra calls to the `checkAccessibility()` method in specific points, like before navigating to a different page and in the `waitModal()` function, so that modal accessibility can be tested too.

We could potentially run this at each test step, but we need to balance the desire to test every component state with the increased execution time resulting by running these checks too often. I tried to add the check function only where I thought it could be useful and I added a boolean flag to the `waitModal` method to allow skipping the check. This is useful in tests where the same modal is opened multiple times. In those cases we can test for the accessibility just the first time a specific modal is opened.

Regarding modals, I also needed to specify to perform the checks only inside the modal (with the `.include()` function), because otherwise the tool was trying to measure the contrast of elements which were behind the modal backdrop, resulting in some false positive errors.

## Notable errors that have been detected.

### Elements must meet minimum color contrast ratio thresholds

> Expected contrast ratio of 4.5:1
> https://dequeuniversity.com/rules/axe/4.11/color-contrast

This was the most common issue. It is the contrast between background and foreground color.

It can be debugged with browser tools. It should show at least AA:

image

Notice that browser tools in this case don't consider opacity, so they could be not reliable for some elements.

Bootstrap provides some [general accessibility](https://getbootstrap.com/docs/5.3/getting-started/accessibility/), but some combinations of colors (e.g. a `code` inside an `.alert-warning`) result in violations of contrast guidelines, so I had to override some CSS values to make it working in our cases.

### Scrollable region must have keyboard access

> Ensure elements that have scrollable content are accessible by keyboard
> https://dequeuniversity.com/rules/axe/4.11/scrollable-region-focusable

This appened for some modals. The user should be able to scroll inside a modal also using the up/down keys and it was not always possible.

The solution is to add a `tabindex="0"` on an element that is inside the modal. In some cases this resulted in a Svelte check warning that I suppressed using

```
<!-- svelte-ignore a11y_no_noninteractive_tabindex -->
```

According to an [open issue](https://github.com/sveltejs/svelte/issues/7953), this is a Svelte check false positive.

### Interactive controls must not be nested

> Ensure interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies
> https://dequeuniversity.com/rules/axe/4.11/nested-interactive

This has been the most complex to solve. You can't have something like a button inside a button, but we need something like that in multiple places, like in the workflow sequence and in the JSON schema form, for actions that are accessible from collapsible header:

image

image

The solution is to not nest the elements, but rather declaring them in sequence and then use CSS to position them one over the other and is not always trivial.

Some suggestions came from this article: https://piccalil.li/blog/accessible-faux-nested-interactive-controls/

### Elements must only use supported ARIA attributes

> Ensure an element's role supports its ARIA attributes

For example, in this case `aria-checked="true"` is allowed only if also adding the `role="checkbox"` attribute:

```html
<span class="text-success boolean-icon" aria-checked="true">
  <i class="bi bi-check"></i>
</span>
```

### Ensure all ARIA attributes have valid values

This happened in some cases where an `aria` attribute was pointing to a non-existing element id.

It happened sometimes also for some slim-select elements with the `aria-activedescendant` attribute. In one case (cda431f434b03a109cb3fbb8f35285ff351d656c) I was able to fix it, but other seemed more trickier. I suspect that this is a slim-select bug rather than an our bug, but I would like to investigate further before opening a issue/PR on slim-select repo.

For the moment I disabled the accessibility checks on the slim-select `.ss-main` elements:

https://github.com/fractal-analytics-platform/fractal-web/blob/29acd43327e76c1e0aac813d41311489c6ad7ce0/tests/base_fixture.js#L28

### Ensure that lists are structured correctly

> `<ul>` and `<ol>` must only directly contain `<li>`, `<script>` or `<template>` elements
> https://dequeuniversity.com/rules/axe/4.11/list

If we need to put something more complex inside a Bootstrap `list-group`, we can just use  `div` instead of `ul` and `li` (ref: af0c1baef0b88190b97f6ca36dd96b401cea560f).

## Manual checks

Some manual checks have been performed using the "Show Tabbing Order" option of browser tools:

image

I also improved the CSS for some elements in the `:focus-visible` state (this is visible only when navigating the page using tabs instead of clicking with the mouse):

image

image

Notice that manual testing with the browser tools displays tons of other warnings. For example, the placeholders in input fields are not considered "visible text labels":

image

I think we should decide what to ignore.

PR #1190, opened 2026-06-22
Enrich admin tasks page

### LINKED ISSUE #1171: Enrich admin/tasks page
Based on https://github.com/fractal-analytics-platform/fractal-web/issues/1165#issuecomment-4708063938

If the admin/tasks page becomes more prominent (because it's the/one entrypoint to mark tasks as core), we may want to enrich its filtering options a bit.

Currently:

Image

---

We could add new filters:
- [x] A filter by owner, with dropdown menu of all user emails - converted to user ID for the API request
- [x] A filter by task-group name (`icontains`)
- [x] Private true/false/unset (default unset)
- [x] Active true/false/unset (default unset)
* ~Resource, with dropdown menu of names - converted to ID for the API request~

And we should also review the table format:
- [x] New column: package name (truncated to 12)
- [x] New column: active
- [x] New column: owner
- [x] New column: group
- [x] "Options" column -> no title, no "info" string, a bit less wide
- [x] If needed: move type into "Info" modal

ychiucco:
closes #1171

## Checklist before merging

### DIFF
--- a/src/routes/v2/admin/tasks/+page.svelte
+++ b/src/routes/v2/admin/tasks/+page.svelte
@@ -2,6 +2,7 @@
 	import { page } from '$app/state';
 	import { arrayToCsv, downloadBlob } from '$lib/common/component_utilities';
 	import { displayStandardErrorAlert, getAlertErrorFromResponse } from '$lib/common/errors';
+	import { sortDropdownUsers } from '$lib/components/admin/user_utilities';
 	import Modal from '$lib/components/common/Modal.svelte';
 	import Paginator from '$lib/components/common/Paginator.svelte';
 	import { normalizePayload, PropertyDescription } from 'fractal-components';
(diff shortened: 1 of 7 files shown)


src/routes/v2/admin/tasks/+page.svelte
```diff
@@ -523,35 +629,23 @@
 							<td>
 								{taskInfo.task.name}
 							</td>
-							<td>{taskInfo.task.version || '-'}</td>
-							<td>{taskInfo.task.type}</td>
-							<td>
-								{taskInfo.relationships.length || '-'}
+							<td title={taskInfo.task.pkg_name}>
+								{truncateString(taskInfo.task.pkg_name, 20)}
```
zonia3000: Just use the `text-truncate` class on `td` element: https://getbootstrap.com/docs/5.2/helpers/text-truncation/

src/routes/v2/admin/tasks/+page.svelte
```diff
@@ -616,6 +710,38 @@
...
+							</button>
+						</h2>
+						<div
+							id="collapse-xy"
+							class="accordion-collapse collapse"
+							data-bs-parent="#accordion-relationships"
+						>
+							<div class="accordion-body p-0">
+								<ul class="list-group noborders">
+									{#each getUsers(selectedTaskInfo) as user (user)}
```
zonia3000: use `user.id` as key for the loop

src/routes/v2/admin/tasks/+page.svelte
```diff
@@ -473,14 +576,17 @@
...
 						<th></th>
 						<th>Name</th>
+						<th>Package</th>
 						<th>Version</th>
-						<th>Type</th>
-						<th># Workflows</th>
+						<th>Active</th>
+						<th>Owner</th>
+						<th>Group</th>
 						<th># Users</th>
```
zonia3000, line 572: Image

I would add `white-space: nowrap;` in the style section (for all the `th` of this table) to prevent the `#` going to a different line.

src/routes/v2/admin/tasks/+page.svelte
```diff
@@ -449,11 +548,15 @@
 					<col width="50" />
 					<col width="30" />
 					<col width="auto" />
```
zonia3000: Probably we should set a width for this column too. Resizing the window will result in the following behavior on the `"auto"` column:

Image

Maybe we can also use a [responsive table](https://getbootstrap.com/docs/5.2/content/tables/#responsive-tables), but we avoided them in the past because it is annoying to use an inner horizontal scrollbar.

zonia3000 (approved):

PR #1191, opened 2026-06-22
Add sharding configuration to end-to-end tests

### LINKED ISSUE #1189: CI error
See https://github.com/fractal-analytics-platform/fractal-web/actions/runs/27937847249/job/82663766575

Image

zonia3000:
## Checklist before merging

Closes #1189

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -10,7 +10,7 @@ permissions: {}
 
 jobs:
   end_to_end_tests:
-    name: 'E2E tests - Node ${{ matrix.node-version }} - ${{ matrix.browser }}'
+    name: 'E2E tests - Node ${{ matrix.node-version }} - ${{ matrix.browser }} ${{ matrix.shard-index }}'
     runs-on: ubuntu-latest
     timeout-minutes: 20
 
@@ -34,6 +34,8 @@ jobs:
       matrix:
         node-version: ['24']
         browser: ['chromium', 'firefox']
+        shard-index: [1, 2, 3]
+        shard-total: [3]
       fail-fast: false
 
     services:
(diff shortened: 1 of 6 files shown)


.github/workflows/end_to_end_tests.yaml
```diff
@@ -74,12 +76,19 @@ jobs:
...
-          name: test-results-folder-node-${{ matrix.node-version }}
+          name: test-results-folder-node-${{ matrix.node-version }}-${{ matrix.browser }}-${{ matrix.shard-index }}
           path: test-results
-          overwrite: true
+
+      - name: Upload blob report to GitHub Actions Artifacts
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: blob-report-${{ matrix.shard-index }}
```
tcompa: Question: shouldn't this filename also include the node version and the browser name?

zonia3000: Yes

zonia3000:
I also added a new job that merges the playwright reports, as suggested in https://playwright.dev/docs/test-sharding

For this PR this is the last result: https://github.com/fractal-analytics-platform/fractal-web/actions/runs/27964459475/artifacts/7797292696

tcompa (approved):

tcompa:
> I also added a new job that merges the playwright reports, as suggested in https://playwright.dev/docs/test-sharding
> 
> For this PR this is the last result: https://github.com/fractal-analytics-platform/fractal-web/actions/runs/27964459475/artifacts/7797292696

Great - would it be meaningful to display this somewhere, rather than only in a downloadable artifact?

An example would be to do as we do for the coverage data on `fractal-server`, with [a banner pointing to a preview hosted on a branch](https://github.com/fractal-analytics-platform/fractal-server/blob/6a1710ebe164ff1ef3e5eb1a063e67fde621304a/README.md?plain=1#L10). Or, similarly, [what we do for some benchmarks](https://github.com/fractal-analytics-platform/fractal-server/blob/6a1710ebe164ff1ef3e5eb1a063e67fde621304a/README.md?plain=1#L12).

What we could do is to add a step after the artifact upload. If this is a push to `main`, we would `git add` and `git commit` the artifact contents to a dedicated branch (e.g. `tests-info`). And then from the README we could add a banner pointing to https://htmlpreview.github.io/?https://github.com/fractal-analytics-platform/fractal-web/blob/tests-info/tests-info/index.html (this was my first test, but the precise branch and relative path are preliminary).

The new GHA step would look a bit like https://github.com/fractal-analytics-platform/fractal-server/blob/6a1710ebe164ff1ef3e5eb1a063e67fde621304a/.github/workflows/benchmarks.yaml#L74-L89.

zonia3000:
> a banner pointing to https://htmlpreview.github.io/?https://github.com/fractal-analytics-platform/fractal-web/blob/tests-info/tests-info/index.html (this was my first test, but the precise branch and relative path are preliminary).

If you try to open a test it doesn't work and displays the HTML source instead. I'm looking for a solution or a different approach.

tcompa:
> If you try to open a test it doesn't work and displays the HTML source instead.

Right. It was working locally (just pointing the browser to `index.html`), but I didn't check the htmlpreview version.

> I'm looking for a solution or a different approach.

Thanks. If it turns out it is too cumbersome, we can also accept not having this feature - although having it available with a single link would make it more valuable than a downloadable artifact.

PR #1197, opened 2026-06-29
Workflow sequence CSS fix and test improvement

### LINKED ISSUE #1193: Review second-line indentation for tasks with long names in workflow page
When tasks have long names, and they never ran, the second line appear with a different indentation 

Image

This seems different from older versions:

Image

I guess this is a consequence of #1188

zonia3000:
## Checklist before merging

Closes #1193


zonia3000:
I've updated the badges:

image

"Documentation" and "End-to-end tests report" are now static badges (without status). As discussed with @tcompa, this is because the badges are not updated when the workflow is executed with the `on: workflow_call` configuration.

PR #1201, opened 2026-06-30
Add User Guide site and help links

### LINKED ISSUE #1196: Fractal user guide
This a first plan for what could become a centralized Fractal user guide/manual - as discussed with @zonia3000.

In terms of contents, our expectations are:
1. Contents is independent on details of a specific instance, even it could mention some potential options. Examples:
    * No: "You should first transfer raw microscope data to the group share at `/shared/filesystem/my-group/"
    * OK: "Depending on the Fractal instance, you may have to stage raw microscope data to a shared filesystem that is accessible to the computing cluster. More details at ... (link to instance-specific wiki/notion page, if relevant)"
2. Contents organization would be `fractal-web`-based, meaning that this is a guide for using the `fractal-web` UI - and not for all possible ways in which Fractal can be used. Examples:
    * OK: Generic instructions ("Here is how to use a workflow template: ...")
    * OK: Generic instructions with some reference of what the backend will do ("Here is how to trigger task collection via pixi (note: if you select `use_lock`, then..)").
    * No: Specific information which are not based on `fractal-web` ("You can automate this by using the Fractal CLI client...")

In terms of organization, we thought about several access routes to useful information:
1. A centrally-hosted version of this guide, hosted at https://fractal-analytics-platform.github.io/user-guide.
2. A user-guide static website available directly for each instance e.g. https://fractal.example.org/help.
3. "Help" buttons distributed at some selected positions throughout the `fractal-web` pages - marked e.g. by a [question mark icon](https://icons.getbootstrap.com/icons/question-circle/). The selected positions would be:
    * Most pages would have a top-level per-page help link (e.g. one on https://fractal.example.org/, one on https://fractal.example.org/v2/projects and one on https://fractal.example.org/v2/projects/4/datasets/5). These buttons would redirect to e.g. https://fractal.example.org/help, https://fractal.example.org/help/projects/ and https://fractal.example.org/help/datasets
    * Some sub-page components may have an additional help message next to them. E.g. when there is a job error we could add a link to https://fractal.example.org/help/job-errors, if relevant.

The reasons for having both 1 and 2 (rather than just 1) are:
* The user guide should have a version that is locked with the fractal-web one. The centralized https://fractal-analytics-platform.github.io/user-guide one may go out-of-sync with respect to the several instances.
* By deploying the user guide on the same domain as `fractal-web`, we will be able to create concrete clickable links to some page (e.g. "See your projects at https://actual-url.org/v2/projects"), rather than generic ones (e.g. "See your projects at https://fractal.example.org/v2/projects").

More technical details:
1. We can start with mkdocs, as a static-site generator (let's keep its configuration simple as we'll replace it with zensical soon enough - or we can actually start with zensical right away).
2. The built site will be build as part of the CI and attached to `fractal-web` releases (with the same version).
3. The build site will be served directly via apache, on a dedicated path of the usual fractal domain (e.g. the `/help` path).
4. `fractal-web` updates will include:
    * A general enable/disable configuration variable (at least for the transitional phase, before this feature becomes stable).
    * Additional help buttons on single pages (either a single one per-page, or possibly additional ones at specific positions).
    * A minimal routing logic that decides which help page should be the target of a given help button.

The current plan is to proceed with a proof of principle of this idea, without any real contents. The steps are:
- Prepare a bunch of mock pages (e.g. "projects", "tasks", "templates") and setup the scaffold for the mkdocs project.
- Customize some fractal-app pages with help buttons linking to these mock pages.
- Add a link to the root of the user manual on the fractal-web home page.

### LINKED ISSUE #1206: Review e2e-tests-artifacts warnings
Image

zonia3000:
## Checklist before merging

Closes #1196
Closes #1206


help/zensical.toml
```diff
@@ -0,0 +1,82 @@
+[project]
+site_name = "Fractal Web User Guide"
+site_url = "https://fractal-analytics-platform.github.io/fractal-web/"
+repo_url = "https://github.com/fractal-analytics-platform/fractal-web"
+repo_name = "fractal-analytics-platform/fractal-web"
+extra_css = ["extra.css"]
+copyright = """
+© Copyright 2026
+Friedrich Miescher Institute for Biomedical Research and
+University of Zurich
```
tcompa: See https://github.com/fractal-analytics-platform/fractal-deployments/issues/637

zonia3000:
I tried to fix the modal not closing when clicking on back button and coming from an external site page.
The main change was 5620ab3304b43fdb4e9c3fb30feba581937262c0, but I reverted it.

I tried to implement something similar to the snippet shown in [SvelteKit Shallow Routing Docs](https://svelte.dev/docs/kit/shallow-routing). It worked for the most trivial cases, but it needed some custom logic to cleanup the extra push state when closing the modal. Moreover, it caused many issues in the cases where the Svelte `goto` function was called from the modal (e.g. when creating a new project or when importing a workflow). In those cases the `history.back()` triggered by closing the modal was interfering with the `goto` that was trying to open a new page. Since both functions are asynchronous it was very difficult to control their interference without over-complicating the `Modal` component code. For this reason I preferred to revert the change.

jluethi:
The docs on how to locally run the user guide has been very helpful, thanks @zonia3000 !

I've written the dataset reference page now so that we can test how we want to handle screenshots and have a realistic reference page.

For screenshot handling, I went with the webp format to save some space. For the full window screenshot, I've resized it to something more reasonable for the eventual display resolution while I kept the tiny zoomed-in screenshots constant.

Before I do this more broadly: Is this screenshot approach fine or should we do it differently?
Automatically generating screenshots would be a nice future goal, but I suspect it to be non-trivial, so I'd start with some realistic ones. Plus, there are some screenshots I want to annotate (like the dataset overview screenshot)

If we have ideas regarding annotation approach / style, we can also discuss those before doing additional reference & concept pages

zonia3000:
I ended up using the same structure for both the versions of the user guide, so that admonition CSS works out of the box also in the minimal version. I then added some CSS to hide header, menu and footer.

image


> Automatically generating screenshots would be a nice future goal, but I suspect it to be non-trivial, so I'd start with some realistic ones. Plus, there are some screenshots I want to annotate (like the dataset overview screenshot)

We could use Playwright to automatically create screenshots, but we need to replicate our end-to-end tests setup. I think it is much easier just to update the images when needed. Moreover, if a screenshot needs to be updated, probably the text content near it needs to be updated too.

I see no trivial way to add annotations to automatically generated screenshots. A completely different approach would be to implement a guided tour of the app with a dedicated library (e.g. https://driverjs.com/), but that needs to be handled separately from these docs pages and it can be complicated too. I did this in the past and I noticed that some users tend to click something different from what the guided action is proposing, often leading to confusing situations in the subsequent steps.

For me the annotations you wrote are clear, but consider that we can also add the help question marks in specific areas of the page, for example:

image

jluethi:
> For me the annotations you wrote are clear, but consider that we can also add the help question marks in specific areas of the page, for example:

Good point! That means we can have smaller reference units. I'd still want to give an overview with the page reference, but we could then link to smaller reference units for details. Especially relevant in the more complex workflow page, while the dataset page is still manageable in complexity

PR #1216, opened 2026-07-07
Add new structure for concepts, reference & snippets

jluethi:
It's really nice to see the technical side of the user guides coming together.

I drafted some content structure and a bit of initial realistic content. It still has lots of placeholders, but I want to explore the key pattern we could use to communicate to our users.

The main ideas are:
1. We separate the content into reference, concepts & snippets. 
2. Reference is what you should see when clicking a given help button. It will contain concrete instructions for how different buttons & functionalities work. It's meant to have screenshots eventually of different elements and extensive help text. Likely with sub-pages to link to e.g. on the workflow page)
3. Concept pages explain more in-depth what things are. These would be what somewhat just navigating the user guide may read or someone wanting to understand how e.g. datasets work
4. Snippets are short summaries of key concept pages. They are embedded in different places (e.g. both in the concept page & in the reference page that then links to the concept page for more)
5. Templates are also available for each sub-page

To play this through, I wrote a full real concept page for the datasets, started adding all the placeholders for what we'll need to explain on the dataset reference page and wrote the snippets for datasets, projects, tasks & workflows. I also added some TODO lists for concepts we'll need to explain on other pages.

@zonia3000 I haven't fully understood how I'd actually make the reference pages I added be the thing that this documentation links to. Could you rewire that? I took over the relevant content examples (e.g. different link structures) from the existing help pages before.

I'm also far form a Zensical expert, this is my first attempt to build that with some Claude help. Feel free to restructure if it's not reasonable 😅


tcompa:
> I'm also far form a Zensical expert, this is my first attempt to build that with some Claude help. Feel free to restructure if it's not reasonable

I'm playing a bit with this, and I don't fully understand what's the intended usage of snippets.

As an example, this would be a concept/datasets page

image

while this would be the reference/dataset one (I zoomed out, so that both the snippet and the main content are visible)

image


Is this expected? That is, is it expected that the "snippet" is such a large block of contents and it is displayed before the reference page contents?

tcompa:
> I'm playing a bit with this, and I don't fully understand what's the intended usage of snippets.

I think now I got it, there are two different `concepts` folders - one with short descriptions and one with longer ones. I will probably rename one of them, for clarity.

tcompa:
> I think now I got it, there are two different `concepts` folders - one with short descriptions and one with longer ones. I will probably rename one of them, for clarity.

This is how a concept page may look when rendering the correct snippet (EDITED, to reflect the original layout):

image

zonia3000:
I've written some info about how to test/develop the user guide (see 1e37fc888bfecfc7536db600b5a30b3ccb122383). I think that in particular it is useful to know the rules about links.

I also updated the help links in fractal-web. I noticed that some markdown syntax was not recognized in the minimal version (the one that is embedded in help modals), so I included some extensions like admonition in zensical configuration for the minimal pages. The HTML is now generated correctly, but it still misses the CSS:

image

image

I'll have a look at how to include it.

jluethi:
@zonia3000 Thanks a lot for adding a user guide writing guide! 

@tcompa Ah yeah, should have named that folder snippets. Thanks for working through this!

If we can make the minimal version a bit less minimal, that'd be nice. Depending on how tricky it is to gradually build up minimal + complexity, we could also see if minimal is "just" the regular page with e.g. hidden side bar, hidden top header and such.

tcompa:
I'm merging this one into `help-pages`, so that we have a single PR for this issue.

PR #1232, opened 2026-07-27
Fix JSchema remove property button misaligned and not working

### LINKED ISSUE #1224: Could `esc` always close the help iframe modal?
It seems to work when you simply open an help dialog and then click `esc`, but if you also browse the user guide from within the iframe and then click `esc` the modal remains there.

### LINKED ISSUE #1231: Delete button on workflow-tasks nested objects mis-aligned
I noticed in fractal-web 1.29.5 that the delete button for Nested Objects is mis-aligned. It's too low and thus almost touched the bottom of the accordion:  

Image

Here in real fractal web example

Image

Also seen in the current sandbox for task3_objects > Nested Object
Image

Another thing I noticed there is that the bottom "Key Add Property" field isn't centered in its box either vertically, not sure if that is intentional

zonia3000:
## Checklist before merging

Closes #1231
Closes #1224

### DIFF
--- a/.github/workflows/end_to_end_tests.yaml
+++ b/.github/workflows/end_to_end_tests.yaml
@@ -59,6 +59,14 @@ jobs:
         with:
           persist-credentials: false
 
+      - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39
+        with:
+          version: '0.11.14'
+          enable-cache: false
+          python-version: '3.12'
+      - run: uv sync --frozen
+      - run: ./build-docs.sh
+
       - name: Set up node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ _Note: Numbers like (\#123) point to closed Pull Requests on the fractal-web rep
 # Unreleased
 
 - Use `text/plain` response in `/profile/token` route (\#1230).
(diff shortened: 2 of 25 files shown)


zonia3000:
Unrelated: in 9de89b620291912d14322622ff858bfdef1a25be I replaced in unit tests many occurrences of `fireEvent` with `@testing-library/user-event`, which is the preferred approach according to `fireEvent` documentation:

> Consider using `@testing-library/user-event` instead, if possible.

The `fireEvent` approach remains used in a couple of places where the preferred approach can't be used.

More details here: https://testing-library.com/docs/user-event/intro/#differences-from-fireevent.

TL;DR:

> `fireEvent` dispatches DOM events, whereas `user-event` simulates full interactions, which may fire multiple events and do additional checks along the way.

zonia3000:
9fad785c3392d320daffae8e2fb065cd6ad61028 fixes this (filter buttons visible even if there are no tasks):

micro-bug-task-filters

zonia3000:
In 138fcadfc936c1577e3e841f7be269080ed51a18 I added an e2e test for the help modals. I was able to test the Escape key behavior but not the back button behavior. It seems that the playwright `page.goBack()` method doesn't do exactly the same of pressing the back button on the browser.

Maybe it is related to the warning at https://playwright.dev/docs/api/class-page#page-go-back

In the test I used the `page.route()` method to [rewrite](https://playwright.dev/docs/mock) the URLs starting with `/help/` and serve the content from the `site` folder. This is necessary to avoid CORS issues. We can't serve the help pages on a different port and we can't rely on vite proxy, which is enable only in dev mode, while the CI uses the production build.

